PR #7, opened 2024-01-17
feat: OKX Provider

davidterpay:


providers/websockets/okx/utils.go
```diff
@@ -0,0 +1,17 @@
...
+	// OKX provides a few different URLs for its Websocket API. The URLs can be found
+	// in the documentation here: https://www.okx.com/docs-v5/en/?shell#overview-production-trading-services
+	// The two production URLs are defined in ProductionURL and ProductionAWSURL. The
+	// DemoURL is used for testing purposes.
+
+	// ProductionURL is the public OKX Websocket URL.
+	ProductionURL = "wss://ws.okx.com:8443/ws/v5/public"
+
+	// ProductionAWSURL is the public OKX Websocket URL hosted on AWS.
+	ProductionAWSURL = "wss://wsaws.okx.com:8443/ws/v5/public"
```
davidterpay, line 13: not used atm, but thought it would be useful to have here.

davidterpay: we can maybe cycle the URLs when it restarts

providers/base/fetch.go
```diff
@@ -101,15 +101,24 @@ func (p *BaseProvider[K, V]) attemptDataUpdate(ctx context.Context, responseCh c
...
+			if err := p.ws.Start(ctx, p.ids, responseCh); err != nil {
+				p.logger.Error("web socket query handler returned error", zap.Error(err))
+			}
 
-	if err := p.ws.Start(ctx, p.ids, responseCh); err != nil {
-		p.logger.Error("web socket query handler returned error", zap.Error(err))
-		return err
+			// If the web socket query handler returns, then the connection was closed. Wait for
+			// a bit before trying to reconnect.
+			time.Sleep(p.cfg.WebSocket.ReconnectionTimeout)
```
Eric-Warehime, line 119: What happens if the remote client doesn't close the connection. I think the read will block indefinitely unless we [SetReadDeadline](https://pkg.go.dev/github.com/gorilla/websocket#Conn.SetReadDeadline).

davidterpay: gonna add this as another thing we need to configure on the web socket providers.

davidterpay: https://linear.app/skip/issue/BLO-758/figure-out-what-the-best-default-dial-configurations-are-for-web

Eric-Warehime (approved):

aljo242 (approved):

PR #15, opened 2024-01-19
feat: Add grpc-gateway reverse proxy on sidecar

Eric-Warehime:
Adds grpc-gateway to the oracle sidecar process so we can just use curl to grab the prices.

Closes BLO-711

Below shows the result. We can now use a single port for serving GRPC requests and also use curl to validate that prices can indeed be retrieved. 

Screenshot 2024-01-18 at 8 50 04 PM


service/server/server.go
```diff
@@ -42,30 +57,60 @@ func NewOracleServer(o servicetypes.Oracle, logger *zap.Logger) *OracleServer {
...
 		}
 	})
 
 	return os
 }
 
+// routeRequest determines if the incoming http request is a grpc or http request and routes to the proper handler
+func (os *OracleServer) routeRequest(w http.ResponseWriter, r *http.Request) {
+	if r.ProtoMajor == 2 && strings.HasPrefix(
+		r.Header.Get("Content-Type"), "application/grpc") {
```
Eric-Warehime, line 73: I took this from the docs https://pkg.go.dev/google.golang.org/grpc#Server.ServeHTTP and it works.

service/server/server.go
```diff
@@ -74,7 +119,7 @@ func (os *OracleServer) StartServer(ctx context.Context, host, port string) erro
 		<-ctx.Done()
 		os.logger.Info("context cancelled, closing oracle")
 
-		os.Close()
+		_ = os.Close()
 		return nil
```
aljo242, line 123: Do we not care about this err?

Eric-Warehime: It's only called when the context is closed and the server is shutting down, so I think that's why it's being thrown out. I've just explicitly ignored it because my editor complains about silently missing it.

We could choose to add a log here I guess. Not sure how noisy that would be--some things like the http.Server always return errors when they shutdown because of context being cancelled I believe.

aljo242 (approved):

PR #20, opened 2024-01-19
refactor: mock generation

aljo242:
Closes BLO-781

Migrate all mocks to use `mockery` and update `make mocks` target to generate easily.
- most files are just regeneration, but some tests were updates to move from `gomock` to `mockery`


providers/base/api/handlers/mocks/api_data_handler.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.30.1. DO NOT EDIT.
+// Code generated by mockery v0.0.0-dev. DO NOT EDIT.
```
Eric-Warehime: Shouldn't these version strings be correct? Are you using a source build of mockery locally or something?

Eric-Warehime (approved):

providers/base/api/handlers/mocks/api_data_handler.go
aljo242: Yeah it was a source build.  I should probably just use the latest version we have in the `go.mod`

PR #23, opened 2024-01-22
chore: Updating Providers with new configuration set up

davidterpay:
Since the API configurations - atomic, name, URL - are now all part of the config we can update the interface to not have to include those. same applies for web sockets.

a bunch of the old config code that was previously specific to each provider is also removed in this PR. these changes overall greatly simplify the process of creating a new provider.


providers/apis/binance/api_handler.go
```diff
@@ -27,26 +26,32 @@ var _ handlers.APIDataHandler[oracletypes.CurrencyPair, *big.Int] = (*APIHandler
 // for more information about the Binance API, refer to the following link:
 // https://github.com/binance/binance-spot-api-docs/blob/master/rest-api.md#public-api-endpoints
 type APIHandler struct {
-	Config
-	BaseURL string
+	// cfg is the config for the Binance API.
+	cfg config.ProviderConfig
+
+	// invertedMarketCfg is convenience struct that contains the inverted market to currency pair mapping.
+	invertedMarketCfg config.InvertedCurrencyPairMarketConfig
```
aljo242: If almost every Handler is going to include the `cfg` and `invertedCfg` why not just contain them in the same struct?

aljo242 (commented):
This is way cleaner.  I think there are some little `lint` errors too.

aljo242 (approved):

PR #24, opened 2024-01-22
chore: Updating app with new config set up + moving code into more a readable structure

davidterpay:


Makefile
```diff
@@ -35,6 +34,9 @@ run-prom-client:
 		-v ./contrib/prometheus/prometheus.yml:/etc/prometheus/prometheus.yml \
 		prom/prometheus
 
+update-local-config:
+	go run ./config/local/main.go
```
aljo242:
```suggestion
	@go run ./config/local/main.go
```

config/local/oracle.toml
```diff
@@ -0,0 +1,342 @@
+update_interval = "1s"
```
aljo242, line 1: so beautiful

service/clients/oracle/metrics_client.go
```diff
@@ -1,33 +1,35 @@
...
 	"cosmossdk.io/log"
+	"google.golang.org/grpc"
 
-	"github.com/skip-mev/slinky/service"
 	"github.com/skip-mev/slinky/service/metrics"
+	"github.com/skip-mev/slinky/service/servers/oracle/types"
 )
 
-var _ service.OracleService = (*MetricsClient)(nil)
+var _ OracleClient = (*MetricsClient)(nil)
```
aljo242: I feel like since the only difference between a metrics client and the normal grpc client is that the metrics are enabled, they should basically just be the same struct with the same routine (with the extra metrics stuff added if enabled.  Its mainly duplicate code

config/local/main.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+const (
+	// localPath is the path to the local config file.
+	localPath = "./config/local/oracle.toml"
+)
+
+// main executes a simple script that encodes the local config file to the local
+// directory.
+func main() {
```
Eric-Warehime: I don't really want people to be running some CLI command to customize and generate their oracle configs. Let's use this as a `go generate` target and avoid outputting a separate binary for this.

config/local/main.go
```diff
@@ -0,0 +1,327 @@
...
+					},
+				},
+			},
+		},
+	},
+}
+
+const (
+	// localPath is the path to the local config file.
+	localPath = "./config/local/oracle.toml"
```
Eric-Warehime: This looks weird--having a relative path to the config file which is based on being executed from the root directory.

go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/cosmos/cosmos-proto v1.0.0-beta.3
 	github.com/cosmos/cosmos-sdk v0.50.3
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
-	github.com/cosmos/gogogateway v1.2.0
+	github.com/cosmos/gogogateway v1.2.0 // indirect
```
Eric-Warehime: You need to merge from main.

go.mod
```diff
@@ -53,6 +53,8 @@ require (
 	mvdan.cc/gofumpt v0.5.0
 )
 
+require github.com/BurntSushi/toml v1.3.2
+
```
Eric-Warehime: Make sure this goes into the main require block.

service/clients/oracle/metrics_client.go
davidterpay: yea agreed, will update

oracle/options.go
```diff
@@ -13,11 +14,22 @@ import (
...
+// WithUpdateInterval sets the update interval on the Oracle.
+func WithUpdateInterval(updateInterval time.Duration) OracleOption {
+	return func(o *OracleImpl) {
+		if updateInterval <= 0 {
+			panic("update interval must be positive")
+		}
+
+		o.updateInterval = updateInterval
+	}
+}
```
Eric-Warehime, line 29: nit: It seems weird to me to have required configurations in here as opposed to the base constructor. But there's probably other options this applies to so no need to deviate just for this one.

service/clients/oracle/metrics_client.go
davidterpay: at second glance this is probably better deferred to a different task, https://linear.app/skip/issue/BLO-802/allow-the-oracle-client-to-continue-to-restart-in-the-case-it-cannot. We are going to want to build in some type of restart logic into the client in case it ever loses its connection.

Makefile
```diff
@@ -35,6 +34,9 @@ run-prom-client:
 		-v ./contrib/prometheus/prometheus.yml:/etc/prometheus/prometheus.yml \
 		prom/prometheus
 
+update-local-config:
+	@go run ./config/local/main.go
+
```
Eric-Warehime: This feels like something we want to just distribute in our artifacts. I guess this is the same comment as I had below which is I prefer to generate these.

Eric-Warehime: Also we should probably have something which ensures that the config files have been regenerated and committed whenever the source is changed, to make sure people don't forget to re-generate.

service/clients/oracle/metrics_client.go
Eric-Warehime: Agree w/ Alex. Maybe using a metrics interface is the easier solution here instead of embedding an OracleClient? That way we can just construct w/ a no-op if not enabled and have exactly the same methods/code across both situations.

service/servers/oracle/server.go
```diff
@@ -57,60 +46,30 @@ func NewOracleServer(o servicetypes.Oracle, logger *zap.Logger) *OracleServer {
...
-// routeRequest determines if the incoming http request is a grpc or http request and routes to the proper handler
-func (os *OracleServer) routeRequest(w http.ResponseWriter, r *http.Request) {
-	if r.ProtoMajor == 2 && strings.HasPrefix(
-		r.Header.Get("Content-Type"), "application/grpc") {
-		os.grpcSrv.ServeHTTP(w, r)
-	} else {
-		os.gatewayMux.ServeHTTP(w, r)
-	}
-}
-
```
Eric-Warehime: Needs rebase--this will remove all the grpc gateway work.

config/local/main.go
davidterpay: updated

config/local/main.go
davidterpay: updated

oracle/options.go
davidterpay: I think as long as we can set sensible defaults we should be good, something i can scope out.

config/gen.go
```diff
@@ -0,0 +1,3 @@
+//go:generate go run ./local/generate_toml.go --oracle-config-path ./local/oracle.toml
+
+package config
```
davidterpay, line 3: we can use this file for staging and build out some CI process to update all of our configs when needed.

aljo242 (approved):

PR #27, opened 2024-01-23
feat: more sidecar metrics

aljo242:
Closes BLO-803

- gauge for prices _per provider_
- gauge for aggregate prices
- re-gen mocks accordingly
- some minor clean up 
- Add `GetResult` interface (alias for `fmt.Stringer`) as a type restriction for our generic code


providers/types/response.go
```diff
@@ -5,22 +5,32 @@ import (
...
 
+// ResponseKey is a type restriction interface for the key of a GetResponse.
+type ResponseKey interface {
+	comparable
+}
+
+// ResponseValue is a type restriction interface for the value of a GetResponse.
+type ResponseValue interface {
+	fmt.Stringer
+}
```
davidterpay, line 16: fire

providers/base/provider.go
```diff
@@ -112,3 +112,15 @@ func (p *BaseProvider[K, V]) GetData() map[K]providertypes.Result[V] {
...
+// Type returns the type of data handler the provider uses
+func (p *Provider[K, V]) Type() providermetrics.ProviderType {
+	switch {
+	case p.apiCfg.Enabled && p.api != nil:
+		return providermetrics.API
+	case p.wsCfg.Enabled && p.ws != nil:
+		return providermetrics.WebSockets
+	default:
+		return "unknown"
+	}
```
davidterpay: this is a small nit but i'd prefer if the provider type (api, websocket, etc). was defined in the provider types i.e. `"github.com/skip-mev/slinky/providers/types"` package instead of `"github.com/skip-mev/slinky/providers/base/metrics"`. think it makes more sense to define all of the provider type variables in a single place.

providers/base/metrics/provider.go
```diff
@@ -17,15 +17,22 @@ const (
...
+	ProviderType string
+)
 
 const (
-	Success Status = "success"
-	Failure Status = "failure"
+	Success    Status       = "success"
+	Failure    Status       = "failure"
+	WebSockets ProviderType = "websockets"
+	API        ProviderType = "api"
```
davidterpay: see other comment, imo i think it makes more sense to move this to the provider types package

providers/base/README.md
```diff
@@ -32,7 +32,7 @@ The `APIDataHandler` interface is primarily responsible for constructing the URL
 // paired with the APIQueryHandler. The APIQueryHandler will use the APIDataHandler
 // to create the URL to be sent to the HTTP client and to parse the response from the
 // API.
-type APIDataHandler[K comparable, V any] interface {
+type APIDataHandler[K comparable, V providertypes.GetResult] interface {
```
davidterpay: nit would update with new K, V

davidterpay: as in the types comparable -> providertypes.ResponseKey etc.

oracle/oracle.go
```diff
@@ -214,14 +215,21 @@ func (o *OracleImpl) fetchPrices(provider providertypes.Provider[oracletypes.Cur
...
 			zap.String("provider", provider.Name()),
+			zap.String("data handler type", string(provider.Type())),
 			zap.String("pair", pair.String()),
 			zap.String("price", result.Value.String()),
 			zap.Duration("diff", diff),
 		)
 		timeFilteredPrices[pair] = result.Value
+
+		// update price metric
+		o.metrics.UpdatePrice(provider.Name(), string(provider.Type()), pair.String(), float64(result.Value.Int64()))
```
davidterpay: hmm, i wonder if it makes more sense to just log this price no matter what like on line 202 right before we filter, and then use the other metric to record whether this pair was used in aggregation. I'd prefer to log the prices as much as possible.

oracle/metrics/server.go
```diff
@@ -29,7 +30,7 @@ type PrometheusServer struct {
 	logger *zap.Logger
 }
 
-// StartPrometheusServer creates a prometheus server if the metrics are enabled and
+// NewPrometheusServer creates a prometheus server if the metrics are enabled and
```
davidterpay: i forgot to delete this server, this is duplicate and should be removed since the new prom server is now in service/servers/prometheus

oracle/metrics/metrics.go
```diff
@@ -24,11 +30,19 @@ type Metrics interface {
 	AddTick()
 	//
 	// TODO: Add more metrics here in later PRs.
+
+	// UpdatePrice price updates the price for the given pairID for the provider.
+	UpdatePrice(name, handlerType, pairID string, price float64)
+
+	// UpdateAggregatePrice rice updates the aggregated price for the given pairID.
+	UpdateAggregatePrice(pairID string, price float64)
 }
```
davidterpay, line 39: I would probably add one more metric here something like 

```
AddPriceIncluded(name, handlertype, pairID, true/false)
```

this way we can track what price feeds are too stale in terms of how often they end up being consumed by the aggregator.

oracle/metrics/server.go
aljo242: Do we need anything in this file?

davidterpay: na

davidterpay: its all copy paste

davidterpay (approved):

oracle/metrics/metrics.go
aljo242: won't the timestamps of the updates being pushed to the db show us how stale they are?

davidterpay: yea

davidterpay: good point just ship

PR #29, opened 2024-01-23
refactor: `CurrencyPair` `String()` method

aljo242:
Closes BLO-805

- use gogoproto annotations so we can implement our own `String()` method and remove the `ToString()` method
- adds `CurrencyPairString()` to construct a currency pair string representation natively
- remove `mocks` from formatting


davidterpay:
much needed

Eric-Warehime (approved):
Nice :D

davidterpay:
I think you just need to update the e2e tests n this should pass

davidterpay (approved):

PR #35, opened 2024-01-24
fix: proto grpc

aljo242:
- fix http get options for some proto grpc gateway options.  They did not include arguments in the URL.
- Fix some proto-lint issues.  We still have some lint issues, but these seem very non-critical to me.
- Refactor `make proto-all` so that `proto-lint` is not called. `proto-lint` always fails so the rest of the targets never run


proto/slinky/incentives/v1/query.proto
```diff
@@ -7,13 +7,14 @@ import "slinky/incentives/v1/genesis.proto";
...
 
+// Query is the query service for the x/incentives module.
 service Query {
   // GetIncentivesByType returns all incentives of a given type. If the type is
   // not registered with the module, an error is returned.
   rpc GetIncentivesByType(GetIncentivesByTypeRequest)
       returns (GetIncentivesByTypeResponse) {
     option (google.api.http).get =
-        "/slinky/incentives/v1/get_incentives_by_type";
+        "/slinky/incentives/v1/get_incentives_by_type/{incentive_type}";
```
aljo242, line 17: this is the actual fix

davidterpay (approved):

PR #36, opened 2024-01-24
fix: oracle client + servers [BLO-804]

g3ndo9kar9:
## In This PR 
- I identify flakiness from `server_test` and remedy it, specifically
   - I implement a blocking feature on the `client` so that tests can be executed with a server that is available, and a client that is fully connected to the server
   - I close the underlying `grpcServer` along with the reverse-proxy to ensure that direct GRPC connections (connections not over HTTP), are handled gracefully on closures


aljo242 (approved):

service/clients/oracle/client.go
```diff
@@ -93,23 +92,51 @@ func NewClient(
...
 	c.logger.Info("starting oracle client", "addr", c.addr)
 
-	conn, err := grpc.Dial(
-		c.addr,
+	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
+
+	if c.blockingDial {
+		opts = append(opts, grpc.WithBlock())
```
Eric-Warehime, line 120: This is discouraged by the grpc library https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md

However, it can make sense to use in test scenarios--as long as we don't include this in production configurations I'm fine with it.

Eric-Warehime (approved):

davidterpay:
This PR is not targeting main

g3ndo9kar9:
> This PR is not targeting main

That is intentional, re the slack msg, this fixes CI failures from a prev. PR

PR #41, opened 2024-01-25
feat: Instrument VoteExtension status [BLO-792]

g3ndo9kar9:


abci/ve/vote_extension.go
```diff
@@ -76,47 +77,51 @@ func (h *VoteExtensionHandler) ExtendVoteHandler() sdk.ExtendVoteHandler {
...
+
+			// measure latency
 			latency := time.Since(start)
 			h.logger.Info(
 				"extend vote handler",
 				"duration (seconds)", latency.Seconds(),
 				"err", err,
 			)
-			h.metrics.ObserveABCIMethodLatency(servicemetrics.ExtendVote, latency)
+			slinkyabci.RecordLatencyAndStatus(h.metrics, latency, err, servicemetrics.ExtendVote)
```
Eric-Warehime, line 99: Is there a reason you're introducing returned errors in this function? The [SDK](https://github.com/cosmos/cosmos-sdk/blob/927ac38dc708bda4efb66cffa643b09697babc38/baseapp/abci.go#L632) catches them and returns `nil` like we are currently doing.

Seems like this will change the behavior of the oracle from defaulting to an empty vote extension on failure to failing to add a vote extension at all. It looks like it will [get swallowed](https://github.com/cometbft/cometbft/blob/635d0b596c7300c98caa57e7aab26a7a4579ab97/internal/consensus/state.go#L2486) by comet, so maybe it nets out to the same outcome in the end?

abci/types/utils.go
```diff
@@ -0,0 +1,25 @@
...
+	"time"
+
+	servicemetrics "github.com/skip-mev/slinky/service/metrics"
+)
+
+// RecordLatencyAndStatus is used by the ABCI handlers to record their e2e latency, and the status of the request
+// to their corresponding metrics objects.
+func RecordLatencyAndStatus(
+	metrics servicemetrics.Metrics, latency time.Duration, err error, method servicemetrics.ABCIMethod,
+) {
```
aljo242, line 13: Should this just be a method on `servicemetrics.Metrics` ?

abci/ve/vote_extension.go
g3ndo9kar9: yes, for each ABCI method, comet's behavior is to eventually panic on errors returned from the app. Hence, the baseapp's implementations in `abci.go` log and ignore errors from our handlers. Hence, I thought it made sense to keep our logic in line w/ our other handlers. For instance, Prepare / ProcessProposal return errors in execution to the base-app despite ultimately being swallowed

Eric-Warehime (approved):

PR #42, opened 2024-01-25
feat: Kucoin WS

davidterpay:
Closes BLO-728


oracle/config/websocket.go
```diff
@@ -140,7 +140,7 @@ func (c *WebSocketConfig) ValidateBasic() error {
 	}
 
 	if c.PingInterval < 0 {
-		return fmt.Errorf("websocket ping interval cannot be negative")
+		return fmt.Errorf("websocket ping interval must be greater than 0")
```
aljo242: We should leave this message as it was.  Since 0 is the valid "disable" value

providers/websockets/kucoin/README.md
```diff
@@ -0,0 +1,13 @@
+# Kucoin Provider
```
aljo242:
```suggestion
# KuCoin Provider
```

KuCoin is always stylized like that. We should probably do the same just to be cute.

providers/websockets/kucoin/messages.go
```diff
@@ -0,0 +1,220 @@
...
+	TickerTopic TopicType = "/market/ticker:"
+)
+
+const (
+	// TickerSubject represents the ticker subject. This should be returned in the
+	// response message when subscribing to the ticker topic.
+	//
+	// ref: https://www.kucoin.com/docs/websocket/spot-trading/public-channels/ticker
+	TickerSubject SubjectType = "trade.ticker"
+)
```
aljo242: let's group all `const` declarations

providers/websockets/kucoin/messages.go
```diff
@@ -0,0 +1,220 @@
...
+//		"type": "ping"
+//	}
+//
+// ref: https://www.kucoin.com/docs/websocket/basic-info/ping
+type PingRequestMessage struct {
+	// ID is the ID of the message.
+	ID string `json:"id"`
+
+	// Type is the type of message.
+	Type string `json:"type"`
```
aljo242: Can embed `BaseMessage` here since they just use the same field

providers/websockets/kucoin/hooks.go
```diff
@@ -0,0 +1,159 @@
+package kucoin
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/goccy/go-json"
```
aljo242: So here we are using this package, but elsewhere we always use the std package.  Is there a reason for this?

davidterpay: that is an accident lol

aljo242 (approved):
lgtm. just have conflicts

PR #43, opened 2024-01-25
feat: extend api and conn handlers

davidterpay:


providers/base/websocket/handlers/ws_conn_handler.go
```diff
@@ -123,3 +140,27 @@ func (h *WebSocketConnHandlerImpl) Close() error {
 
 	return h.conn.Close()
 }
+
+// SetConnection is used to set the connection to the data provider.
+func (h *WebSocketConnHandlerImpl) SetConnection(conn *websocket.Conn) {
+	h.Lock()
+	defer h.Unlock()
+
+	h.conn = conn
```
aljo242: Since we currently only allow for one connection, should we error or return if the connection is already set?

aljo242 (approved):

PR #49, opened 2024-01-26
feat: huobi ws provider

aljo242:
Closes BLO-736


providers/websockets/huobi/ws_data_handler.go
```diff
@@ -0,0 +1,153 @@
...
+		subscriptionResponse SubscriptionResponse
+		tickerStream         TickerStream
+	)
+
+	reader, err := gzip.NewReader(bytes.NewReader(message))
+	if err != nil {
+		h.logger.Error("error creating gzip reader", zap.Error(err))
+		return resp, nil, err
+	}
+	defer reader.Close()
```
davidterpay, line 78: does the defer need to happen before the error?

providers/websockets/huobi/ws_data_handler.go
```diff
@@ -0,0 +1,153 @@
...
+		h.logger.Error("error reading gzip", zap.Error(err))
+		return resp, nil, err
+	}
+
+	// attempt to unmarshal to ping and check if field values are not nil
+	if err := json.Unmarshal(uncompressed.Bytes(), &pingMessage); err == nil && pingMessage.Ping != 0 {
+		h.logger.Debug("received ping message")
+		updateMessage, err := NewPongMessage(pingMessage)
+		return resp, updateMessage, err
+	}
```
davidterpay, line 92: should we add a test case for this?

davidterpay (commented):
lgtm, would add the test case for pongs

providers/websockets/huobi/ws_data_handler.go
aljo242: If there is an error, the reader that will be returned is `nil`

davidterpay (approved):

PR #50, opened 2024-01-26
chore: Fix dockerfile

Eric-Warehime:
Run the oracle sidecar in docker.


config/local/oracle.toml
```diff
@@ -19,7 +19,7 @@ production = false
     name = ""
     read_buffer_size = 0
     write_buffer_size = 0
-    handshake_timeout = "0s"
```
davidterpay: Not that this will be over-written any time the oracle is started as it refreshes the local config.

davidterpay:
why is the handshake timeout affecting the API providers? Do you have any logs we can look at?

davidterpay (approved):

Eric-Warehime:
> why is the handshake timeout affecting the API providers? Do you have any logs we can look at?

False alarm--seems like it was just staleness. Dockerfile changes are all that's needed to get it running properly.

PR #52, opened 2024-01-27
feat: Implement `oracle_prices` metric in PreBlock [BLO-790]

g3ndo9kar9:
## In This PR
- I introduce the `oracle_prices` metric
- The metric is implemented as a gauge, and reports the `big.Int` value reported as a float64 (in scientific notation, as opposed to an integer)
     - I also test that things don't break when using 256 bit integers


oracle/oracle.go
```diff
@@ -199,8 +199,10 @@ func (o *OracleImpl) fetchPrices(provider providertypes.Provider[oracletypes.Cur
 
 	timeFilteredPrices := make(map[oracletypes.CurrencyPair]*big.Int)
 	for pair, result := range prices {
+		floatValue, _ := result.Value.Float64() // we ignore the accuracy in this conversion
+
 		// update price metric
-		o.metrics.UpdatePrice(provider.Name(), string(provider.Type()), pair.String(), float64(result.Value.Int64()))
+		o.metrics.UpdatePrice(provider.Name(), string(provider.Type()), pair.String(), floatValue)
```
g3ndo9kar9, line 206: context: I make this change as the `.Int64()` method on big.Int's behavior is undefined for non-int64 values, hence we can report incorrect information in the common-case

aljo242 (approved):

PR #65, opened 2024-01-31
feat: close web socket after maximum number of read errors is reached

davidterpay:
ATM we rely on gorilla web socket to panic after 1000 consecutive read errors are reached, this means that there is a corrupted connection to the provider. This PR adds a configurable number read err count.


providers/websockets/bitstamp/utils.go
```diff
@@ -30,6 +30,7 @@ var (
 		WriteTimeout:        config.DefaultWriteTimeout,
 		ReadTimeout:         config.DefaultReadTimeout,
 		PingInterval:        10 * time.Second,
+		MaxReadErrorCount:   config.DefaultReadBufferSize,
```
aljo242:
```suggestion
		MaxReadErrorCount:   config.DefaultMaxReadErrorCount,
```

Wrong value used throughout

aljo242 (changes requested):

providers/websockets/bitstamp/utils.go
davidterpay: shit

davidterpay: good call

aljo242 (approved):

PR #67, opened 2024-01-31
feat: multi-connection websocket providers

aljo242:
Add `MaxSubscriptionsPerConnection` to config

Closes BLO-815


oracle/config/websocket.go
```diff
@@ -42,6 +42,10 @@ const (
 	// DefaultPingInterval is the default interval at which the provider should send
 	// ping messages to the server.
 	DefaultPingInterval = 0 * time.Second
+
+	// DefaultMaxSubscriptionsPerConnection is the default maximum subscriptions
+	// an provider can handle per-connection.
```
davidterpay: nit: can you explain what happens when this value is 0? Does it just establish a single connection?

providers/base/fetch.go
```diff
@@ -110,8 +110,37 @@ func (p *Provider[K, V]) startWebSocket(ctx context.Context, responseCh chan<- p
 			return ctx.Err()
 		default:
 			p.logger.Debug("starting websocket query handler")
-			if err := p.ws.Start(ctx, p.ids, responseCh); err != nil {
-				p.logger.Error("websocket query handler returned error", zap.Error(err))
+			// create sub handlers
```
davidterpay: note that `p.ws.Start` will block the current thread/process - so this won't actually spin up multiple web socket connections. imo i think it makes more sense to first make the slices and then call `startWebSocket` with each slice in seperate go routines. `startWebSocket` should just function as it currently does but only has to deal with a subset of IDs.

davidterpay (commented):
can you add a test case that see's that it correctly splits into two connections and correctly shuts down the go routines.

davidterpay (approved):

PR #69, opened 2024-01-31
feat: add `Decimals` field to `CurrencyPair`

aljo242:
Closes BLO-859


x/oracle/types/currency_pair.go
```diff
@@ -51,22 +58,14 @@ func CurrencyPairFromString(s string) (CurrencyPair, error) {
 		return CurrencyPair{}, fmt.Errorf("incorrectly formatted CurrencyPair: %s", s)
 	}
 	cp := CurrencyPair{
-		Base:  strings.ToUpper(split[0]),
-		Quote: strings.ToUpper(split[1]),
+		Base:     strings.ToUpper(split[0]),
```
davidterpay, line 104: I think in this case the string should be something like base/quote/decimals

pkg/math/math.go
```diff
@@ -21,7 +21,7 @@ func Min[V int | int64 | uint64 | int32 | uint32](vals ...V) V {
 }
 
 // Float64StringToBigInt converts a float64 string to a big.Int.
-func Float64StringToBigInt(s string, decimals int) (*big.Int, error) {
+func Float64StringToBigInt(s string, decimals int64) (*big.Int, error) {
```
g3ndo9kar9, line 23: Do we lose precision here? It looks like we're parsing the returned float value into a float64, and converting that float64 value into a big.Float. We shld j convert directly

x/oracle/types/currency_pair.go
```diff
@@ -6,14 +6,16 @@ import (
...
 
 const (
-	ethereum = "ETHEREUM"
+	DefaultDecimals  = 8
+	EthereumDecimals = 18
 )
 
 // NewCurrencyPair returns a new CurrencyPair with the given base and quote strings.
-func NewCurrencyPair(base, quote string) CurrencyPair {
+func NewCurrencyPair(base, quote string, decimals int64) CurrencyPair {
```
g3ndo9kar9, line 15: We should enforce a uniqueness constraint in the x/oracle schema on the base/quote pairs of currency-pairs?

aljo242: Yes!

one question I had is that, since we are doing this, should we let people query using both 

`BASE/QUOTE/DEC`
and
`BASE/QUOTE` 

? 

IMO i think simplifying to just use `BASE/QUOTE` for the query would be good, but down to discuss.

pkg/math/math.go
aljo242: Wanna make an issue for this?

g3ndo9kar9: https://linear.app/skip/issue/BLO-860/dont-convert-price-to-float64-in-floattostring

x/oracle/types/currency_pair.go
```diff
@@ -31,48 +34,88 @@ func (cp *CurrencyPair) ValidateBasic() error {
...
 
-// String returns a string representation of the CurrencyPair, in the following form "ETH/BTC".
+// String returns a string representation of the CurrencyPair, in the following form "ETH/BTC/8" where 8 is the number of decimals.
 func (cp CurrencyPair) String() string {
+	return fmt.Sprintf("%s/%s/%d", cp.Base, cp.Quote, cp.Decimals)
+}
+
+// ID returns a string representation of the CurrencyPair, in the following form "ETH/BTC" where 8 is the number of decimals.
+// This function differs from String() which also includes the decimal information.
+func (cp *CurrencyPair) ID() string {
```
davidterpay: Maybe ticker is better than ID?

x/oracle/types/currency_pair.go
```diff
@@ -31,48 +34,88 @@ func (cp *CurrencyPair) ValidateBasic() error {
...
+	decimals, err := strconv.ParseInt(split[2], 10, 64)
+	if err != nil {
+		return CurrencyPair{}, fmt.Errorf("incorrectly formatted CurrencyPair: %s", s)
+	}
+
 	cp := CurrencyPair{
-		Base:  strings.ToUpper(split[0]),
-		Quote: strings.ToUpper(split[1]),
+		Base:     strings.ToUpper(split[0]),
+		Quote:    strings.ToUpper(split[1]),
```
davidterpay, line 90: upper casing here introduces weird consequence where the base and quote are not correctly formatted but will not return an error.

x/oracle/types/currency_pair.go
aljo242: Yeah, this is changed in https://github.com/skip-mev/slinky/pull/79 which stacks on top of this

x/oracle/types/currency_pair.go
```diff
@@ -31,48 +34,88 @@ func (cp *CurrencyPair) ValidateBasic() error {
...
+		split := strings.Split(cpStr, "/")
+		if len(split) == 2 {
+			return cpStr, nil
+		}
+
+		return "", fmt.Errorf("invalid string provided: %w", err)
+	}
+
+	return cp.ID(), nil
+}
```
davidterpay: IMO this should just return an error in the case when decimals is not included, feels like a footgun otherwise.

davidterpay (commented):
can u confirm that the oracle runs with these changes

PR #74, opened 2024-02-01
feat: factor vote-aggregation into it's own pkg [BLO-814]

g3ndo9kar9:
## In This PR
- I factor the logic for aggregating votes from an extended commit to the `strategies/aggregator` pkg
    - I introduce a `VoteAggregator` abstraction to handle this -> we can export this into the dydx PrepareProposal
- I also move all constants used in ABCI into a `slinky/abci/types` pkg


abci/strategies/aggregator/vote_aggregator.go
```diff
@@ -0,0 +1,214 @@
...
+			dva.logger.Debug(
+				"failed to decode price",
+				"currency_pair_id", cpID,
+				"err", err,
+			)
+
+			// If the price cannot be decoded, continue.
+			//
+			// TODO: Should we return an error here instead?
+			continue
```
aljo242, line 194: is this a carryover or a new TODO?

aljo242 (commented):
super clean !

abci/strategies/aggregator/vote_aggregator.go
g3ndo9kar9: ya this is a carry-over

g3ndo9kar9: Would you prefer that we remove the TODO or change the behavior?

g3ndo9kar9: Linear: https://linear.app/skip/issue/BLO-865/resolve-potential-errors-in-preblock-when-aggregating-votes-prices

aljo242 (approved):

davidterpay (approved):

PR #75, opened 2024-02-01
refactor: Factor vote-weighted median logic out of preblock and into its own pkg [BLO-814]

g3ndo9kar9:
## In This PR
- I factor out the `vote-weighted` median calculation from preblock into it's own package


pkg/math/vote-weighted/math.go
```diff
@@ -1,4 +1,4 @@
-package math
+package voteweighted
```
aljo242: So it looks like we have a discrepancy between the package name `voteweighted` and the dir name `vote-weighted`. 

I say we just use `voteweighted`.

Also the file `math.go` should probably be `voteweighted.go`

aljo242 (approved):

davidterpay (approved):

PR #78, opened 2024-02-02
refactor: float conversion methods

aljo242:
Closes BLO-860

- introduce `BigFloatToBigInt` which is used by `Float64StringToBigInt`
- this means that float64 strings are not converted in a way where any precision is lost
- add more tests for the package 
- some formatting changes from running `make format`


davidterpay (approved):

pkg/math/math.go
```diff
@@ -11,40 +10,44 @@ func Min[V int | int64 | uint64 | int32 | uint32](vals ...V) V {
...
 		return nil, err
 	}
 
-	return Float64ToBigInt(floatNum, decimals), nil
+	return BigFloatToBigInt(bigFloat, decimals), nil
 }
 
 // Float64ToBigInt converts a float64 to a big.Int.
-//
-// TODO: Is there a better approach to this?
```
g3ndo9kar9, line 35: Where is this being used?

pkg/math/math.go
```diff
@@ -11,40 +10,44 @@ func Min[V int | int64 | uint64 | int32 | uint32](vals ...V) V {
...
+
+// BigFloatToBigInt converts a big.Float to a big.Int.
+func BigFloatToBigInt(f *big.Float, decimals int) *big.Int {
+	bigFloat := new(big.Float)
 	factor := big.NewInt(1).Exp(big.NewInt(10), big.NewInt(int64(decimals)), nil)
-	coin.SetInt(factor)
+	bigFloat.SetInt(factor)
 
-	bigval.Mul(bigval, coin)
+	f.Mul(f, bigFloat)
```
g3ndo9kar9, line 47: Maybe add a benchmark or smth? Given how frequently this is used, we shld at least have some introspection into perf

g3ndo9kar9 (approved):
👑

pkg/math/math.go
aljo242: added benchmarks and new `make test-bench` target

PR #79, opened 2024-02-02
refactor: cp keepers and queries

aljo242:
Use `CurrencyPair.Ticker()` for all querying and msg server interactions.

This PR also refactors some internal `x/oracle` keeper methods us the `Ticker()` as input, simplifying some of their logic.

Also includes a bunch of cleanup found along the way


proto/slinky/oracle/v1/query.proto
```diff
@@ -36,14 +36,8 @@ message GetAllCurrencyPairsResponse {
...
-  // CurrencyPairSelector represents the types that the user may provide to the
-  // request to identify a CurrencyPair
-  oneof currency_pair_selector {
-    // CurrencyPair type
-    CurrencyPair currency_pair = 1;
-    // string representation of the CurrencyPair
-    string currency_pair_id = 2;
-  }
+  // CurrencyPairID is the string representation of the CurrencyPair.
+  string currency_pair_id = 2;
```
aljo242, line 40: Query refactor removes the use of `oneof` and moves to just always query by CurrencyPairID ("ticker")

x/oracle/keeper/grpc_query.go
```diff
@@ -42,93 +42,51 @@ func (q queryServer) GetAllCurrencyPairs(ctx context.Context, _ *types.GetAllCur
...
-		// fail if any other type of CurrencyPairSelector is given
-		return nil, fmt.Errorf("invalid CurrencyPairSelector given in request (consult documentation)")
-	}
 
 	// unwrap ctx
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	// get the QuotePrice + nonce for the given CurrencyPair
-	qpn, err := q.k.GetPriceWithNonceForCurrencyPair(ctx, cp)
+	cps, err := q.k.currencyPairs.Get(ctx, req.CurrencyPairId)
```
aljo242, line 54: Since we are only querying by CurrencyPairID, we can simplify the keeper accesses to a single read where we get the `CurrencyPairState`

x/oracle/keeper/grpc_query.go
```diff
@@ -42,93 +42,51 @@ func (q queryServer) GetAllCurrencyPairs(ctx context.Context, _ *types.GetAllCur
...
-			return nil, fmt.Errorf("error unmarshalling CurrencyPairID: %v", err)
-		}
-
+	for _, cpID := range req.CurrencyPairIds {
 		// unwrap ctx
 		ctx := sdk.UnwrapSDKContext(goCtx)
 
-		// get the QuotePrice + nonce for the given CurrencyPair
-		qpn, err := q.k.GetPriceWithNonceForCurrencyPair(ctx, cp)
+		cps, err := q.k.currencyPairs.Get(ctx, cpID)
```
aljo242, line 80: Similarly, here we can simplify by getting the `CurrencyPairState` directly

x/oracle/keeper/grpc_query.go
```diff
@@ -42,93 +42,51 @@ func (q queryServer) GetAllCurrencyPairs(ctx context.Context, _ *types.GetAllCur
...
 
 		prices = append(prices, types.GetPriceResponse{
-			Price:    &qpn.QuotePrice,
-			Nonce:    qpn.Nonce(),
-			Decimals: cp.Decimals,
-			Id:       id,
+			Price:    cps.Price,
+			Nonce:    cps.Nonce,
+			Decimals: cps.Decimals,
+			Id:       cps.Id,
```
aljo242, line 89: `CurrencyPairState` includes all of the information we need

x/oracle/keeper/keeper.go
```diff
@@ -98,13 +98,13 @@ func NewKeeper(
...
-	k.currencyPairs.Remove(ctx, cp.ID())
+func (k Keeper) RemoveCurrencyPair(ctx sdk.Context, cpID string) {
+	k.currencyPairs.Remove(ctx, cpID)
 }
 
 // HasCurrencyPair returns true if a given CurrencyPair is stored in state, false otherwise.
-func (k Keeper) HasCurrencyPair(ctx sdk.Context, cp types.CurrencyPair) bool {
-	ok, err := k.currencyPairs.Has(ctx, cp.ID())
+func (k Keeper) HasCurrencyPair(ctx sdk.Context, cpID string) bool {
+	ok, err := k.currencyPairs.Has(ctx, cpID)
```
aljo242, line 107: Moving `keeper` methods to use `cpID` instead of a `cp` simplifies the logic for the callers of these functions.  Before, a `cp` would need to be constructed, only for its `String()` or `Ticker()` representation to be used internally

proto/slinky/oracle/v1/query.proto
davidterpay: maybe `Ticker` makes more sense than `CurrencyPairID`?

davidterpay: gets confusing when considering that there is a separate ID for each CP

proto/slinky/oracle/v1/query.proto
```diff
@@ -63,10 +57,15 @@ message GetPriceResponse {
 
 // GetPricesRequest takes an identifier for the CurrencyPair
 // in the format base/quote
-message GetPricesRequest { repeated string currency_pair_ids = 1; }
+message GetPricesRequest {
+  // CurrencyPairIDs is a list of the string representations of the
+  // CurrencyPairs.
+  repeated string currency_pair_ids = 1;
```
davidterpay, line 63: same as above except now its `Tickers`

PR #80, opened 2024-02-05
chore: turbo lint

aljo242:
Closes BLO-870

add a bunch of linters
- make according fixes
- I tried to label in the `.golangci.yaml` file what each linter does

Types of changes:
- always use "%w" in `fmt.Errorf` calls where we are wrapping errors
- always put "." at the end of Godoc comment
- preallocate arrays where we can
- use `errors.As()` instead of "==" when checking errors
- add `t.Helper()` and `t.Parallel()` where appropriate


x/sla/keeper/price_feeds_test.go
```diff
@@ -344,6 +344,8 @@ func (s *KeeperTestSuite) TestRemovePriceFeeds() {
 }
 
 func checkEquality(t *testing.T, sla1, sla2 slatypes.PriceFeed) {
+	t.Helper()
```
davidterpay, line 347: wut do this?

abci/proposals/proposals.go
```diff
@@ -198,7 +199,7 @@ func (h *ProposalHandler) PrepareProposalHandler() sdk.PrepareProposalHandler {
 // injectTx will only appear once regardless of how many times you attempt to inject it.
 // If injectTx is large enough, all originalTxs may end up being excluded from the returned tx array.
 func (h *ProposalHandler) injectAndResize(appTxs [][]byte, injectTx []byte, maxSizeBytes int64) [][]byte {
-	var returnedTxs [][]byte
+	returnedTxs := make([][]byte, 0, len(appTxs)) // init with initial capacity
```
davidterpay: This is going to keep appending to the empty list instead. I dont think this needs to be preallocated?

davidterpay (changes requested):
there is one functional change in this PR that should be changed, otherwise.,...turbo lint gud

x/sla/keeper/price_feeds_test.go
aljo242: this just designate that it is a helper (visually) and this also changes how this function is logged when running verbose tests

davidterpay (approved):

PR #87, opened 2024-02-06
feat: Conversion Aggregation Function

davidterpay:
closes BLO-873


pkg/math/oracle/median.go
```diff
@@ -0,0 +1,252 @@
...
+				zap.Error(err),
+				zap.Any("conversions", conversion),
+			)
+
+			continue
+		}
+
+		// Calculate the converted price.
+		convertedPrice, err := m.CalculateConvertedPrice(cp, conversion, medians)
+		if err != nil {
```
aljo242: we should def add test cases for this codepath

pkg/math/oracle/utils.go
```diff
@@ -0,0 +1,74 @@
+package oracle
+
+import (
+	"fmt"
+	"math/big"
+)
+
+// ScaledDecimals is the standard number of decimal places each price will be converted to
+// during the conversion process.
+const ScaledDecimals = 40
```
aljo242: docs earlier said this was 36?

aggregator/math.go
```diff
@@ -43,31 +43,39 @@ func ComputeMedian() AggregateFn[string, map[types.CurrencyPair]*big.Int] {
...
+	}
+}
 
-				medianPrices[cp] = medianPrice
-			} else {
-				medianPrices[cp] = prices[middle]
-			}
+// CalculateMedian calculates the median from a list of big.Ints. Returns an
+// average if the number of values is even.
+func CalculateMedian(values []*big.Int) *big.Int {
```
aljo242, line 55: should this live in `pkg/math` as well?

oracle/config/aggregate_market.go
```diff
@@ -49,6 +49,17 @@ type FeedConfig struct {
...
 
+// AggregateFeedConfig represents all of the conversion markets that can be used to convert the
+// price of a currency pair to a common currency pair.
+type AggregateFeedConfig struct {
+	// CurrencyPair is the currency pair that the oracle will convert to.
+	CurrencyPair oracletypes.CurrencyPair `mapstructure:"currency_pair" toml:"currency_pair"`
+
+	// Conversions is a list of conversion operations that will be used to convert the price
+	// of the currency pair to the common currency pair.
+	Conversions [][]Conversion `mapstructure:"operations" toml:"operations"`
```
aljo242: Maybe lets make a type alias:

```go
type Conversions []Conversion
```

pkg/math/oracle/median.go
davidterpay: the `MedianAggregator` should never be constructed if a bad path is included. Ill add some for posterity.

pkg/math/oracle/utils.go
davidterpay: will update.

aggregator/math.go
davidterpay: Im fine keeping it in the other since it otherwise is going to duplicate testing, can add a test case to move at a later time.

aljo242 (approved):

PR #89, opened 2024-02-07
feat: Update providers with latest conversion market changes

davidterpay:
the oracle is now consuming 233 price feeds across 9 providers.


cmd/oracle/main.go
```diff
@@ -67,11 +67,18 @@ func main() {
 		return
 	}
 
+	// Create the conversion market aggregator.
```
davidterpay, line 70: Swapping the aggregation function for the conversion one.

config/local/generate_toml.go
```diff
@@ -29,245 +29,473 @@ import (
...
+		},
+		Production: true,
+		// -----------------------------------------------------------	//
+		// ----------------------Metrics Config-----------------------	//
+		// -----------------------------------------------------------	//
+		Metrics: config.MetricsConfig{
+			Enabled:                 true,
+			PrometheusServerAddress: "0.0.0.0:8002",
+		},
+		UpdateInterval: 1500 * time.Millisecond,
```
davidterpay, line 57: I increased this from 1s to 1.5s since each price feed (btc/usdt, btc/usdc, etc.) no longer point directly to btc/usd. We should do some more analysis to see how much this new conversion logic interferes with the frequency at which we can expected an aggregated price update for a given price feed.

pkg/math/oracle/median.go
```diff
@@ -59,7 +59,7 @@ func (m *MedianAggregator) AggregateFn() aggregator.AggregateFn[string, map[orac
 	) map[oracletypes.CurrencyPair]*big.Int {
 		// Calculate the median price for each price feed.
 		feedMedians := aggregator.ComputeMedian()(feedsPerProvider)
-		m.logger.Debug("calculated median prices for raw price feeds", zap.Any("num_prices", len(feedMedians)))
+		m.logger.Info("calculated median prices for raw price feeds", zap.Any("num_prices", len(feedMedians)))
```
```suggestion
		m.logger.Info("calculated median prices for raw price feeds", zap.Int("num_prices", len(feedMedians)))
```

providers/apis/README.md
```diff
@@ -8,6 +8,22 @@ API providers utilize rest APIs to retrieve data from external sources. The data
...
+    * Check if a given market is supported:
+        * `curl https://api.binance.com/api/v3/ticker/price?symbol=BTCUSDT | jq`
 * [Coinbase](./coinbase/README.md) - Coinbase is a cryptocurrency exchange that provides a free API for fetching cryptocurrency data. Coinbase is a **primary data source** for the oracle.
-* [CoinGecko](./coingecko/README.md) - CoinGecko is a cryptocurrency data aggregator that provides a free API for fetching cryptocurrency data. CoinGecko is a **secondary data source** for the oracle.
+    * Check all supported markets: 
+        * `curl https://api.exchange.coinbase.com/currencies | jq`
+        * `curl https://api.exchange.coinbase.com/products | jq`
+    * Check if a given market is supported: 
+        * `curl https://api.coinbase.com/v2/prices/{DYDX-USDC}/spot | jq`
+* [CoinGecko](./coingecko/README.md) - CoinGecko is a cryptocurrency data aggregator that provides a free API for fetching cryptocurrency data. CoinGecko is a **secondary data source** for the oracle. This is not recommended for use in production.
```
aljo242, line 25: Should we just make a note that we will deprecate this provider?  Or make plans to remove altogether?  Is there a good reason for keeping?

aljo242 (approved):
just some small nits and a question

providers/apis/README.md
davidterpay: I think we should make a separate prod config and let this be in the local/dev one.

PR #90, opened 2024-02-07
feat: add options + retaining oracle data in proposals [BLO-840]

g3ndo9kar9:
## In This PR
- I add a field `retainOracleDataInWrappedHandler` which determines whether or not the ProposalHandler shld pass the oracle-data to the wrapped Prepare / ProcessProposalHandler. 
- Previously, our behavior was to inject on Prepare, and not inject in Process -> this led to inconsistencies in dydx / berachain


abci/proposals/options.go
```diff
@@ -0,0 +1,12 @@
+package proposals
+
+// Option is a function that enables optional configuration of the ProposalHandler.
+type Option func(*ProposalHandler)
+
+// RetainOracleDataInWrappedProposalHandler returns an Option that configs the ProposalHandler
+// to pass the injected extend-commit-info to the wrapped proposal handler.
+func RetainOracleDataInWrappedProposalHandler() Option {
```
aljo242:
```suggestion
// RetainOracleDataInWrappedProposalHandler returns an Option that configures the
// ProposalHandler to pass the injected extend-commit-info to the wrapped proposal handler.
func RetainOracleDataInWrappedProposalHandler() Option {
```

aljo242 (approved):

PR #97, opened 2024-02-08
feat: Market Configurations Proto

davidterpay:
closes BLO-876


proto/slinky/marketmap/v1/market.proto
```diff
@@ -0,0 +1,109 @@
...
+option go_package = "github.com/skip-mev/slinky/x/marketmap/types";
+
+import "gogoproto/gogo.proto";
+
+// Ticker represents a price feed for a given asset pair i.e. BTC/USD. The price feed
+// is scaled to a number of decimal places and has a minimum number of providers required
+// to consider the ticker valid.
+message Ticker {
+    // ID is a unique sequentially-generated value for the ticker.
+    uint64 id = 1;
```
davidterpay: this should have a 1:1 mapping with what is stored in the `x/oracle` module.

aljo242: Do we actually need this?  Since the kvstore access in `x/oracle` is `BASE/QUOTE` doesnt using that already give use 1:1 mapping?

aljo242 (approved):
Lgtm

aljo242:
> Lgtm

Just run make proto-all before merging

PR #114, opened 2024-02-11
feat: Add marketmap module/keeper

Eric-Warehime:
Adds marketmap module and keeper implementation.

I've adhered generally to the proto implementations that were committed to the feature branch, although I think a broader conversation should be had about how those are laid out (we can change that in a separate PR if we agree on a different data layout there).

I've written this initial implementation around the idea of 2 access patterns:
* Reasoning about which providers to run and the sets of tickers each provider needs to retrieve
* Reasoning about which tickers to return in the oracle's Prices API and how to calculate each of those prices

Managing them separately seems the most reasonable to me since it enables us to have a 2 phase operation in which we turn on new currency pair feeds in the MarketConfigs and subsequently enable the prices in the x/oracle module by creating an aggregation config.

I've left out validation for now--happy to add that logic and tests in if it's crucial for this PR, I think a lot of it already exists in the proto types package already.

BLO-879


x/marketmap/keeper/keeper.go
```diff
@@ -0,0 +1,116 @@
...
+
+	slinkytypes "github.com/skip-mev/slinky/pkg/types"
+	"github.com/skip-mev/slinky/x/marketmap/types"
+)
+
+type Keeper struct {
+	cdc codec.BinaryCodec
+
+	// keeper dependencies
+	oracleKeeper types.OracleKeeper
```
aljo242: I think we really should not be adding anything from `x/oracle` as a dependency into `x/marketmap`.  My thinking was for `x/marketmap` to expose [hooks](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/types/hooks.go#L121) in a pattern similar to how some other cosmos sdk modules do. 

Something like `AfterMarketMapUpdate()` which can be used by `x/oracle` (or `x/prices`) to get a view of changes without adding a direct keeper dependency

x/marketmap/keeper/keeper.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, err
+	}
+	configs, err := iter.Values()
+	if err != nil {
+		return nil, err
+	}
+	return configs, err
+}
+
+func (k Keeper) GetAllAggregationConfigs(ctx sdk.Context) ([]types.PathsConfig, error) {
```
aljo242, line 56: nit: let's Godoc all exported functions

x/marketmap/keeper/keeper.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// module authority
+	authority sdk.AccAddress
+
+	// collections
+	// marketConfigs is keyed by provider name and provides the MarketConfig for each given provider
+	marketConfigs collections.Map[string, types.MarketConfig]
+	// aggregationConfigs is keyed by CurrencyPair string (BASE/QUOTE) and contains the PathsConfig used
+	// to do price aggregation for a given canonical Ticker
+	aggregationConfigs collections.Map[string, types.PathsConfig]
```
aljo242: This is not in the scope of this PR, but I'm wondering if we could/should introduce a type alias

```go
type TickerString string
```

so we can more easily reason about the types of keys we are using to index into maps throughout the codebase

x/marketmap/keeper/keeper.go
g3ndo9kar9: Could get around this via an interface + implementation for the x/oracle module?

x/marketmap/keeper/keeper.go
g3ndo9kar9: is it not possible to use the CurrencyPair obj. as they key to this Map?

Eric-Warehime: Yeah that seems like a great idea. 🚀

x/marketmap/keeper/keeper.go
Eric-Warehime: Removing oracleKeeper interface from deps here. Plan is to have a follow up PR which includes hooks.

x/marketmap/types/keys.go
```diff
@@ -0,0 +1,35 @@
...
+	TickerStringCodec = codec.NewStringKeyCodec[TickerString]()
+)
+
+// MarketProvider is the unique name used to key the MarketConfigs in the marketmap module.
+// It is identical to the MarketConfig.Name property which is stored as the value in the Keeper.marketConfigs map.
+type MarketProvider string
+
+// TickerString is the key used to identify unique pairs of Base/Quote with corresponding PathsConfig objects--or in other words AggregationConfigs.
+// The TickerString is identical to Slinky's CurrencyPair.String() output in that it is `Base` and `Quote` joined by `/` i.e. `$BASE/$QUOTE`.
+type TickerString string
```
aljo242, line 35: does it make sense to put these in `./pkg/types` as common types?

aljo242 (approved):
lgtm.

Regarding my other comment, maybe we can open an issue / discussion on this.  Not blocking moving forward.

Eric-Warehime:
> lgtm.
> 
> Regarding my other comment, maybe we can open an issue / discussion on this. Not blocking moving forward.

Done--BLO-904

PR #117, opened 2024-02-12
feat: Config updater for the base provider

davidterpay:
closes BLO-901

This PR introduces a `ConfigUpdater` interface that allows the base provider to be updated asynchronously. Anytime a configuration update is received, this will trigger the provider to restart with the new configurations.


providers/base/config.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"go.uber.org/zap"
+)
+
+// listenConfigUpdater listens for updates from the config updater and updates the
+// provider's internal configurations. This will trigger the provider to restart
+// and is blocking until the context is cancelled.
+func (p *Provider[K, V]) listenConfigUpdater(ctx context.Context) {
```
aljo242:
```suggestion
func (p *Provider[K, V]) listenOnConfigUpdater(ctx context.Context) {
```

providers/base/provider.go
```diff
@@ -90,15 +97,66 @@ func (p *Provider[K, V]) Start(ctx context.Context) error {
...
+		select {
+		case <-p.restartCh:
+			// If any of the provider's configurations have changed, the provider will
+			// be signalled to restart.
+			p.logger.Info("restarting provider")
+			cancel()
+
+			// Wait for the fetch loop to stop.
+			err := <-errCh
+			p.logger.Debug("provider fetch loop stopped", zap.Error(err))
```
aljo242, line 132:
```suggestion
Restart:
	for {
		// Create a new context for the fetch loop. This allows us to cancel the fetch loop
		// when the provider needs to be restarted.
		fetchCtx, cancel := context.WithCancel(ctx)
		defer cancel()

		// Start the fetch loop.
		errCh := make(chan error)
		go func() {
			errCh <- p.fetch(fetchCtx)
		}()

		select {
		case <-p.restartCh:
			// If any of the provider's configurations have changed, the provider will
			// be signalled to restart.
			p.logger.Info("restarting provider")
			cancel()

			// Wait for the fetch loop to stop.
			err := <-errCh
			p.logger.Debug("provider fetch loop stopped", zap.Error(err))
			break Restart
```

I think this code is a bit hard to reason about, so adding this label (while redundant) makes things a bit clearer.

providers/base/fetch.go
```diff
@@ -80,7 +76,8 @@ func (p *Provider[K, V]) startAPI(ctx context.Context, responseCh chan<- provide
 // attemptAPIDataUpdate tries to update data by fetching and parsing API data.
 // It logs any errors encountered during the process.
 func (p *Provider[K, V]) attemptAPIDataUpdate(ctx context.Context, responseCh chan<- providertypes.GetResponse[K, V]) {
-	if len(p.ids) == 0 {
+	ids := p.GetIDs()
```
g3ndo9kar9, line 79: Unclear what p.GetIDs() is?

providers/base/utils.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	providertypes "github.com/skip-mev/slinky/providers/types"
+)
+
+type (
+	// The ConfigUpdater is an interface that be utilized to fetch the configuration
+	// for a given provider. This allows the provider to be updated asynchronously.
+	ConfigUpdater[K providertypes.ResponseKey] interface {
+		// GetIDs is the channel that is used to update the set of IDs that the provider
+		// will fetch data for. This blocks until there is a viable update.
```
g3ndo9kar9, line 12: Any reason to choose a channel v. a RWMutex here? Don't you want all Providers to read the updated ids concurrently? Do we expect a ConfigUpdater per Provider?

providers/base/provider.go
```diff
@@ -90,15 +97,66 @@ func (p *Provider[K, V]) Start(ctx context.Context) error {
...
+		case err := <-errCh:
+			return err
+		case <-ctx.Done():
+			return <-errCh
+		}
```
g3ndo9kar9, line 143: Why do we need to kill the providers if there is a config update? Don't we j need to update the set of IDs it's fetching for?

aljo242: It works cleanly for that with REST providers, but for websockets, things get more complicated.  We would need to do unsubscribe/subscribe logic based on the diff we get in providers. 

This just simplifies things, creating a new connection so we have a fresh set of subscriptions.

providers/base/utils.go
aljo242: Right now, this is being done per provider

providers/base/fetch.go
davidterpay: IDs are effectively currency pairs, the base provider just calls them IDs to have some generality.

providers/base/utils.go
davidterpay: yea the interface that the config client, which is the client that is responsible for updating all providers, will use is the ConfigUpdater - which is specific per provider.

providers/base/provider.go
```diff
@@ -90,15 +97,78 @@ func (p *Provider[K, V]) Start(ctx context.Context) error {
...
+		select {
+		case <-p.restartCh:
+			// If any of the provider's configurations have changed, the provider will
+			// be signalled to restart.
+			p.logger.Info("restarting provider")
+			cancel()
+
+			// Wait for the fetch loop to stop.
+			err := <-errCh
+			p.logger.Debug("provider fetch loop stopped", zap.Error(err))
```
aljo242, line 132:
```suggestion
			p.logger.Debug("provider fetch loop stopped", zap.Error(err))
			continue MainLoop
```

aljo242 (approved):

PR #118, opened 2024-02-12
feat: manually stopping providers

davidterpay:
closes BLO-902


providers/base/provider.go
```diff
@@ -139,12 +148,33 @@ MainLoop:
 			// to exit when the context is cancelled.
 			retErr = <-errCh
 			break MainLoop
+		case <-p.stopCh:
+			// If the provider is manually stopped, we stop the fetch loop and return.
+			p.logger.Debug("stopping provider")
+			cancel()
+			retErr = <-errCh
+			break MainLoop
```
aljo242, line 157: Using `break MainLoop` here and in the other cases makes sense.  Maybe we add `continue MainLoop` for the restart case?

this might be overkill

aljo242 (approved):
lgtm with small idea

PR #120, opened 2024-02-13
feat: Add query flags to query commands for all modules

### LINKED ISSUE #119: Missing query flags
Currently, Slinky CLI commands doesn't query flags. For example, attempting to run the command:
```
initiad query oracle currency-pairs --node <INITIA_RPC_URL>
```

Results in the following error message:
```sh
Error: unknown flag: --node
Usage:
  initiad query oracle currency-pairs [flags]
```
The problem seems to be from in the CLI commands defined in the various Slinky module ﻿query.go files – such as [this one](https://github.com/skip-mev/slinky/blob/f1b58a03250e301b54467a5e5563f024e58ef221/x/oracle/client/cli/query.go). These definitions currently fail to support query flags that are added through ﻿flags.AddQueryFlagsToCmd(cmd).

h/t @beer-1 for identifying the issue/necessary changes

Eric-Warehime:
Resolve #119 

Adds the standard global query flags to all of the modules' query commands.

### DIFF
--- a/x/incentives/client/cli/query.go
+++ b/x/incentives/client/cli/query.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


aljo242 (approved):
legend

PR #128, opened 2024-02-14
feat: add `metadata` field to `Ticker` and `MarketConfig`

aljo242:
Closes BLO-913

Adds metadata fields to msg index 15 for both messages.  This gives room for the messages to be extended if necessary, while using  a field index that still [takes one byte to encode](https://protobuf.dev/programming-guides/proto3/#assigning).

Add `IsValid()` function to new `pkg/json` package.  This checks the basic validity of if a byte array is json.

The metadata fields are chosen as `string` and not `bytes` in the proto, because strings can be compared as map keys.


aljo242:
One other thing to note is that we probably should choose some length limit for the JSON strings.  I'm not sure what that limit would look like and would like to discuss this.

Eric-Warehime (approved):
LGTM. In terms of adding limits to the json size...since it's a string I guess it's slightly harder to reason about encoding size? Maybe just saying 1K and seeing if we ever need to break that is reasonable. Since the updates are currently constrained by the permissioned actor and in the future might have some permissionless element to them, I don't think not validating this poses and huge risk.

PR #129, opened 2024-02-14
feat: add `Params` to `x/marketmap`

aljo242:
Closes BLO-917

introduces `Params` with `MarketAuthority` field.  

This allows for the v1 functionality of having a single authority control updates which is configurable.

Since this is configurable, this allows for governance to change this authority.

Adds grpc query for the object.


x/marketmap/types/params.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+)
+
+// DefaultParams returns default incentive parameters.
```
Eric-Warehime:
```suggestion
// DefaultParams returns default marketmap parameters.
```

Eric-Warehime (approved):
LGTM. We need still need to integrate validation of the market authority in the market update calls I think? Also in a separate PR we should add the module versioning data to the Params as well.

PR #130, opened 2024-02-14
chore: Updating CP to Tickers

davidterpay:


Eric-Warehime (commented):
I think this PR could stand to be broken down into potentially quite a few separate PRs.

It seems like you're
* migrating from CurrencyPair to Ticker in a lot of places
* deprecating large portions of the config for the sidecar?? (but the oracle.toml file isn't updated)
* Converting the oracle config from toml to json

It's hard for me to see the full vision of where this refactoring is going and why we need certain parts of it.

davidterpay:
yea, im definitely breaking all of this up, just having this branch be here so that i can reference later on.

PR #132, opened 2024-02-15
refactor: remove market config from oracle and providers

davidterpay:
NOTE: This PR is breaking and must be merged into `feat/marketmap` after all other PRs that branch on this are merged here.

* since we are using the marketmap module to determine the set of providers and tickers they are responsible for, i am removing the separate oracle config that exists rn.
* moving the file to be json based instead of toml based since the marketmap types are json as well (default of proto)


oracle/config/api.go
```diff
@@ -8,29 +8,29 @@ import (
 // APIConfig defines a config for an API based data provider.
 type APIConfig struct {
 	// Enabled is a flag that indicates whether the provider is API based.
-	Enabled bool `mapstructure:"enabled" toml:"enabled"`
+	Enabled bool `mapstructure:"enabled" json:"enabled"`
```
aljo242: Do we need `mapstructure` anymore?

davidterpay: we do not i believe

davidterpay: removed

aljo242:
tests failing

davidterpay:
tests will keep failing till all other changes are merged into this branch

oracle/config/websocket.go
```diff
@@ -59,63 +59,63 @@ const (
 // WebSocketConfig defines a config for a websocket based data provider.
 type WebSocketConfig struct {
 	// Enabled is a flag that indicates whether the provider is websocket based.
-	Enabled bool `mapstructure:"enabled" toml:"enabled"`
+	Enabled bool `json:"enabled"`
```
Eric-Warehime, line 62: Wasn't the `Enabled` field mostly put in place to work around peculiarities w/ generating configs via toml? I'm thinking it doesn't make sense to have now--a provider will be enabled if there are feeds for it to fetch and disabled if not.

oracle/config/oracle.go
```diff
@@ -13,22 +13,17 @@ import (
...
-	UpdateInterval time.Duration `mapstructure:"update_interval" toml:"update_interval"`
+	UpdateInterval time.Duration `json:"updateInterval"`
 
 	// Providers is the list of providers that the oracle will fetch prices from.
-	Providers []ProviderConfig `mapstructure:"providers" toml:"providers"`
-
-	// Market defines the market configurations for how currency pairs will be resolved to a
-	// final price. Each currency pair can have a list of convertable markets that will be used
-	// to convert the price of the currency pair to a common currency pair.
-	Market AggregateMarketConfig `mapstructure:"market" toml:"market"`
```
Eric-Warehime, line 24: How is the aggregation logic going to be configured now without AggregateMarketConfig?

davidterpay: going to reference the marketmap module, which will have the relevant market information.

aljo242 (approved):

PR #134, opened 2024-02-15
chore: Updating Binance with ticker type

davidterpay:


pkg/math/math.go
```diff
@@ -20,7 +20,7 @@ func Min[V int | int64 | uint64 | int32 | uint32](vals ...V) V {
 }
 
 // Float64StringToBigInt converts a float64 string to a big.Int.
-func Float64StringToBigInt(s string, decimals int) (*big.Int, error) {
+func Float64StringToBigInt(s string, decimals uint64) (*big.Int, error) {
```
davidterpay, line 23: switching this to uint64 since decimals are uint64

providers/apis/binance/api_handler.go
```diff
@@ -10,115 +10,125 @@ import (
...
+		_, resolvedOk := resolved[ticker]
+		_, unresolvedOk := unresolved[ticker]
+
+		if !resolvedOk && !unresolvedOk {
+			unresolved[ticker] = fmt.Errorf("no response")
+		}
 	}
 
-	return providertypes.NewGetResponse[slinkytypes.CurrencyPair, *big.Int](resolved, unresolved)
+	return providertypes.NewGetResponse[mmtypes.Ticker, *big.Int](resolved, unresolved)
```
aljo242: Could also be `constants.NewPriceGetResponse`

providers/apis/binance/api_handler_test.go
```diff
@@ -176,142 +126,117 @@ func TestCreateURL_US(t *testing.T) {
...
 			name: "valid single",
-			cps:  []slinkytypes.CurrencyPair{slinkytypes.NewCurrencyPair("BITCOIN", "USDT")},
+			cps:  []mmtypes.Ticker{constants.BITCOIN_USDT},
 			response: testutils.CreateResponseFromJSON(
 				`[{"symbol":"BTCUSDT","price":"46707.03000000"}]`,
 			),
 			expected: providertypes.NewGetResponse(
-				map[slinkytypes.CurrencyPair]providertypes.Result[*big.Int]{
-					slinkytypes.NewCurrencyPair("BITCOIN", "USDT"): {
+				map[mmtypes.Ticker]providertypes.Result[*big.Int]{
```
aljo242, line 138: `providertypes.Result[*big.Int]` -> `constants.PriceResult`?

providers/constants/oracle.go
```diff
@@ -0,0 +1,58 @@
...
+	// requests and parse the response into price data.
+	PriceAPIDataHandler = apihandlers.APIDataHandler[mmtypes.Ticker, *big.Int]
+)
+
+var (
+	// The following are the default tickers for the oracle sidecar.
+	//
+	// USD denominated tickers.
+	ATOM_USD     = mmtypes.NewTicker("ATOM", "USD", 8, 1)
+	AVAX_USD     = mmtypes.NewTicker("AVAX", "USD", 8, 1)
```
aljo242, line 29: Want to suppress this warning from the linter?  I like having constants all caps like this.  We could exclude this rule from the golangci config

aljo242:
lint just needs `make format` to be run

providers/apis/binance/api_handler.go
aljo242: also idk if `constants` is a good package name for these type aliases.

I think `constants` makes sense for the ticker constants, but the generic type aliases still make sense in `providertypes` to me

davidterpay: going to add
```
	// PriceResponse is a type alias for the response from the price providers.
	PriceResponse = providertypes.GetResponse[mmtypes.Ticker, *big.Int]
```

providers/constants/oracle.go
```diff
@@ -0,0 +1,62 @@
...
+	PriceResponse = providertypes.GetResponse[mmtypes.Ticker, *big.Int]
+)
+
+var (
+	// The following are the default tickers for the oracle sidecar.
+	//
+	// USD denominated tickers.
+	ATOM_USD     = mmtypes.NewTicker("ATOM", "USD", 8, 1)
+	AVAX_USD     = mmtypes.NewTicker("AVAX", "USD", 8, 1)
+	BITCOIN_USD  = mmtypes.NewTicker("BITCOIN", "USD", 8, 1)
```
Eric-Warehime, line 30: Some feedback we've gotten from Initia is that they currently have `ETHEREUM` and `BITCOIN` on chain and want `ETH` `BTC`. I guess we should just make a conscious decision to support 1 or the other in terms of long/short forms and not have both. Because right now we have short forms on Berachain and long on Initia.

I'm fine w/ long form names, but then you'll want to standardize things like `AVALANCHE` instead of `AVAX`.

PR #145, opened 2024-02-19
refactor: market map config refactor

aljo242:


proto/slinky/marketmap/v1/market.proto
```diff
@@ -24,89 +32,27 @@ message Ticker {
...
+  // prices of a set of tickers to a common ticker.
+  repeated Path paths = 5 [ (gogoproto.nullable) = false ];
+
+  // Providers is the list of provider configurations for the given ticker.
+  repeated ProviderConfig providers = 6 [ (gogoproto.nullable) = false ];
+
   // MetadataJSON is a string of JSON that encodes any extra configuration
   // for the given ticker.
   string metadata_JSON = 15;
 }
```
davidterpay, line 45:
```suggestion

  // Paths is the list of convertable markets that will be used to convert the
  // prices of a set of tickers to a common ticker.
  repeated Path paths = 5 [ (gogoproto.nullable) = false ];

  // Providers is the list of provider configurations for the given ticker.
  repeated ProviderConfig providers = 6 [ (gogoproto.nullable) = false ];

  // MetadataJSON is a string of JSON that encodes any extra configuration
  // for the given ticker.
  string metadata_JSON = 7;
}
```

x/marketmap/types/ticker.go
```diff
@@ -14,6 +14,8 @@ const (
 	// DefaultMinProviderCount is the minimum number of providers required for a
 	// ticker to be considered valid.
 	DefaultMinProviderCount = 1
+	// MaxPathLength is the maximum length of a path for a ticker conversion.
+	MaxPathLength = 2
```
davidterpay, line 18: Why are we setting this at 2? I think 3 should be allowed

davidterpay (approved):

Eric-Warehime (approved):

PR #152, opened 2024-02-20
refactor: the final config

aljo242:
Closes BLO-923


Eric-Warehime (approved):
Shipping to get us unblocked--needs followup w/ tests, todo fixes, refactoring of Ticker type to remove non-comparable fields, etc.

x/marketmap/keeper/keeper.go
```diff
@@ -59,22 +70,151 @@ func (k Keeper) GetAllTickers(ctx sdk.Context) ([]types.Ticker, error) {
...
+	iter, err := k.providers.Iterate(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+	providers, err := iter.Values()
+	if err != nil {
+		return nil, err
+	}
+	return providers, err
+}
```
davidterpay: This will return duplicates if a provider is configured to retrieve multiple tickers. should this be a distinct check instead?

davidterpay: oh i see this returns the values associated with each currency pair.

aljo242: yeah this should just be removed.  The `Map` functions should only be used here.

x/marketmap/keeper/keeper.go
```diff
@@ -59,22 +70,125 @@ func (k Keeper) GetAllTickers(ctx sdk.Context) ([]types.Ticker, error) {
...
+	// Check if MarketConfig already exists for the provider
+	alreadyExists, err := k.paths.Has(ctx, types.TickerString(ticker.String()))
 	if err != nil {
 		return err
 	}
+	if alreadyExists {
+		return types.NewTickerAlreadyExistsError(types.TickerString(ticker.String()))
+	}
+	// Create the config
+	return k.paths.Set(ctx, types.TickerString(ticker.String()), paths)
```
davidterpay, line 175: Should this do a basic check to make sure that the tickers supplied in a path are in the main set of tickers, similar to what is done in validate basic on the marketmap?

davidterpay: Not critical rn, can be added later.

davidterpay (approved):
lgtm, left a small comment on checking something that can be pushed in a later PR, ideally after i get unblocked.

PR #156, opened 2024-02-20
feat: msg-server v2

aljo242:
Add basic msg server with `MsgUpdateMarketMap`
- add check to `keeper.ValidateState` that ensures all currency pairs included in paths are already stored in state (unless  the path is the currency pair itself).  This function can be called after all batched updates to the market map to verify the validity of the resulting state.

`MsgUpdateMarketMap` will be the main transaction that can interact with the module.  Currently is has the field `CreateMarkets` which is the list of markets to be created.  Later, this can be extended in a compatible way to add `UpdateMarkets` and `DeleteMarkets` if desired, each with their own logic.

will need to be updated with a message for setting params.


proto/slinky/marketmap/v1/tx.proto
```diff
@@ -0,0 +1,53 @@
...
+// MsgUpdateMarketMap defines a message carrying a payload for updating the
+// x/marketmap module.
+message MsgUpdateMarketMap {
+  option (cosmos.msg.v1.signer) = "signer";
+  option (amino.name) = "slinky/x/marketmap/MsgCreateMarkets";
+
+  option (gogoproto.equal) = false;
+
+  // Signer is the signer of this transaction (notice, this may not always be a
+  // node from the SecondTier)
```
davidterpay: what is the second tier?

x/marketmap/keeper/keeper.go
```diff
@@ -192,3 +194,32 @@ func (k *Keeper) CreateMarket(ctx sdk.Context, ticker types.Ticker, paths types.
 
 	return k.SetLastUpdated(ctx)
 }
+
+// ValidateState is called after keeper modifications have been made to the market map to verify that
+// the aggregate of all updates has led to a valid state.
+func (k *Keeper) ValidateState(ctx sdk.Context, creates []types.CreateMarket) error {
```
davidterpay, line 222: small nit, is there a reason this has to take in a list of creates instead of a single one?

x/marketmap/keeper/msg_server.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/skip-mev/slinky/x/marketmap/types"
+)
+
+// msgServer is the default implementation of the x/marketmap MsgService.
+type msgServer struct {
+	k Keeper
+}
+
+// NewMsgServer returns the default implementation of the x/marketmap message service.
+// NewMsgServer returns the default implementation of the x/marketmap message service.
```
davidterpay:
```suggestion
```

x/marketmap/keeper/msg_server.go
```diff
@@ -0,0 +1,64 @@
...
+
+var _ types.MsgServer = (*msgServer)(nil)
+
+// UpdateMarketMap updates the marketmap from the given message.
+func (ms msgServer) UpdateMarketMap(goCtx context.Context, msg *types.MsgUpdateMarketMap) (*types.MsgUpdateMarketMapResponse, error) {
+	if msg == nil {
+		return nil, fmt.Errorf("unable to process nil msg")
+	}
+
+	// Update the module's parameters.
```
davidterpay:
```suggestion
```

x/marketmap/keeper/msg_server.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// create markets
+	for _, market := range msg.CreateMarkets {
+		err := ms.k.CreateMarket(ctx, market.Ticker, market.Paths, market.Providers)
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO: call creation hooks
+	}
```
davidterpay: I would add a small godoc on why we have to create the markets before we validate state.

x/marketmap/types/msg.go
```diff
@@ -0,0 +1,63 @@
...
+
+		if len(market.Paths.Paths) == 0 {
+			return fmt.Errorf("at least one path is required for a ticker to be calculated")
+		}
+
+		for _, path := range market.Paths.Paths {
+			if err := path.ValidateBasic(); err != nil {
+				return err
+			}
+		}
```
davidterpay: I thought we decided that the paths were optional?

x/marketmap/keeper/keeper.go
aljo242: The `UpdateMarketMap` tx contains a list of creates (so it's sort of batching all of these changes).  We need to verify that for all of these changes, the state is valid

x/marketmap/types/msg.go
```diff
@@ -0,0 +1,63 @@
...
+		for _, provider := range market.Providers.Providers {
+			// check for duplicate providers
+			if _, seen := seenProviders[provider.Name]; seen {
+				return fmt.Errorf("duplicate provider found: %s", provider.Name)
+			}
+			seenProviders[provider.Name] = struct{}{}
+
+			if provider.OffChainTicker == "" {
+				return fmt.Errorf("got empty off chain ticker for provider %s", provider.Name)
+			}
```
davidterpay: nit: I would just call validate basic on the `ProviderConfig` while doing the duplicate name check

x/marketmap/types/msg.go
aljo242: yeah I wrote this before we had your changes merged - will update

davidterpay (approved):

PR #159, opened 2024-02-21
fix: remove url validation

beer-1:
Found slinky currently accepts only `localhost:8080` not `0.0.0.0:8080`.

`url.ParseRequestURI`, in your AppConfig Validation, only accepts `localhost:8080` or `tcp://0.0.0.0:8080`.
But, in grpc client, `tcp://0.0.0.0:8080` is not working, so only option we can use is `localhost:8080` for now.

By simply removing the validation, we can make oracle server to accept `0.0.0.0:8080`.

or you can implement regex to check "IP:PORT".


davidterpay (commented):
I think we should still do at least a len check on the URL. otherwise lgtm. probably why its failing CI rn

oracle/config/app.go
```diff
@@ -82,10 +81,6 @@ func (c *AppConfig) ValidateBasic() error {
 		return nil
 	}
```
aljo242, line 83:
```suggestion

if len(c.OracleAddress) == 0 {
     return fmt.Errorf("oracle address must not be empty")
}
```

Eric-Warehime (approved):

PR #167, opened 2024-02-22
feat: Introduce `MaxCacheAge` + `MaxPriceAge`  to oracle / config [BLO-934]

g3ndo9kar9:
## In This PR
 - I introduce the `maxCacheAge` parameter
      - Previously, the `updateInterval` served the purpose of refreshing the oracle's price-cache + removing entries that had been in cache longer than this parameter
      - This is unsatisfactory, as the behavior of some providers is to not return updates unless a change has occurred, in which case, we unnecessarily pruned prices
- The `maxCacheAge` allows this parameter to be set independently of the `UpdateInterval` so that price-freshness + the duration of entries in the cache are set independently


aljo242 (approved):

providers/websockets/gate/ws_data_handler.go
```diff
@@ -111,7 +111,7 @@ func (h *WebsocketDataHandler) CreateMessages(
 		market, ok := h.cfg.Market.CurrencyPairToMarketConfigs[cp.String()]
 		if !ok {
 			h.logger.Debug("market not found for currency pair", zap.String("currency_pair", cp.String()))
-			return nil, fmt.Errorf("market not found for currency pair: %s", cp.String())
+			continue
```
g3ndo9kar9, line 120: Ignoring this error is the same behavior as (okx, coinbase, kraken, mexc, cryptodotcom) this provider failed to start without ignoring this error, investigation is logged [here](https://linear.app/skip/issue/BLO-932/investigate-currencypairtomarketconfigs-parse-error-in-createmessages)

aljo242 (approved):
LGTM - just lint failing

PR #168, opened 2024-02-22
feat: Remove un-necessary go routines

davidterpay:


config/local/market.json
```diff
@@ -48,153 +48,153 @@
       "decimals": 8,
       "min_provider_count": 1
     },
-    "BITCOIN/USD": {
+    "BTC/USD": {
```
davidterpay, line 51: Updating the provider supported markets based on things that got no responses.

oracle/oracle.go
```diff
@@ -271,18 +271,15 @@ func (o *OracleImpl) setLastSyncTime(t time.Time) {
...
-	}()
+	for cp, price := range prices {
+		floatValue, _ := price.Float64() // we ignore the accuracy in this conversion
+
+		o.metrics.UpdateAggregatePrice(
+			strings.ToLower(cp.String()),
+			cp.Decimals,
+			floatValue,
+		)
+	}
```
aljo242, line 282: do we want to eventually flag this behavior?  I feel like in certain prod scenarios ppl might wanna not run this.

aljo242 (approved):

oracle/oracle.go
davidterpay: this is a no op if metrics is disabled, what type of flagging did you have in mind?

PR #172, opened 2024-02-23
feat: Config Client Provider

davidterpay:


proto/slinky/marketmap/v1/query.proto
```diff
@@ -47,6 +47,9 @@ message GetMarketMapResponse {
   // Version is the schema version for the MarketMap data structure and query
   // response.
   uint64 version = 3;
+
+  // ChainId is the chain identifier for the market map.
+  string chain_id = 4;
```
davidterpay, line 52: Adding a ChainID to the response so that we can uniquely associate marketmaps with chains.

x/marketmap/types/market.go
```diff
@@ -62,3 +62,8 @@ func (mm *MarketMap) ValidateBasic() error {
 
 	return nil
 }
+
+// String returns the string representation of the market map.
+func (mm MarketMap) String() string {
```
aljo242, line 67:
```suggestion
func (mm *MarketMap) String() string {
```

aljo242 (approved):
Overall LGTM

What about the case in the future where we may want to enable the super-sidecar?

I think this design may limit our ability to support multiple market maps for multiple chainIDs.  Maybe there is a way around this tho. wdyt?

davidterpay:
Yea, i purposefully built it to fetch only a single chain ID just for simplicity, but we can imagine it being relatively easy to extend this implementation to fetch multiple chain IDs - ultimately taking a superset of the providers and currency pairs that need to be supported across the various chains by the super-side car.

PR #178, opened 2024-02-26
test: cleanup `x/marketmap` and test genesis functions

aljo242:


x/marketmap/types/genesis.go
```diff
@@ -1,23 +1,31 @@
...
 }
 
 // ValidateBasic performs basic validation on the GenesisState.
 func (gs *GenesisState) ValidateBasic() error {
-	return gs.MarketMap.ValidateBasic()
+	if err := gs.MarketMap.ValidateBasic(); err != nil {
+		return err
+	}
+
+	if gs.LastUpdated < 0 {
```
Eric-Warehime: If we don't want negatives can't we just use a uint?

x/marketmap/types/provider_test.go
```diff
@@ -1,3 +1,33 @@
 package types_test
 
-// TODO finish
```
Eric-Warehime, line 3: Inspirational!

Eric-Warehime (approved):

x/marketmap/types/genesis.go
aljo242: Yeah.  But the sdk block height type is an `int64` (extremely annoying) so I just went with consistency with that.  I suppose we will never lose anything if we make it a `uint64` tho

PR #179, opened 2024-02-26
feat(ci): add dev environment

Zygimantass:
# In this PR

Adds a build and deploy step that generates Docker images for:
1. Standard sidecar build that other people might use
2. Standard simapp build without config generation
3. The same pair as above but with configs in the Docker container (slinky-testapp and slinky-e2e-sidecar)

The chain and sidecar reset their states on restart right now, so if that is undesireable, we might be able to think of another solution. (the concern here is how to make it easy for devs to reset state on demand).


docker-compose.yml
```diff
@@ -8,14 +8,14 @@ services:
...
     build:
       context: .
-      dockerfile: contrib/images/slinky.e2e.oracle.Dockerfile
-    volumes:
-      - ./config/local/oracle.json:/oracle/oracle.json
-      - ./config/local/market.json:/oracle/market.json
+      dockerfile: contrib/images/slinky.sidecar.e2e.Dockerfile
+    # volumes:
+    #   - ./config/local/oracle.json:/oracle/oracle.json
+    #   - ./config/local/market.json:/oracle/market.json
```
Eric-Warehime: Just remove this instead of commenting it out?

Makefile
```diff
@@ -127,15 +130,15 @@ $(BUILD_DIR)/:
 	mkdir -p $(BUILD_DIR)/
 
 # build-configs builds a slinky simulation application binary in the build folder (/test/.slinkyd)
-build-configs: build-test-app
```
Eric-Warehime, line 130: Why?

Makefile
```diff
@@ -262,3 +265,16 @@ format:
...
 .PHONY: format
+
+###############################################################################
+###                                dev-deploy                               ###
+###############################################################################
+
+deploy-dev:
+	touch ${LEVANT_VAR_FILE}
+	yq e -i '.sidecar_image |= "${SIDECAR_IMAGE}"' ${LEVANT_VAR_FILE}
+	yq e -i '.chain_image |= "${CHAIN_IMAGE}"' ${LEVANT_VAR_FILE}
```
Eric-Warehime, line 276: `yq` is already installed on the github workflow image where this will run?

Eric-Warehime (approved):

PR #190, opened 2024-02-27
fix: ensure parity between `market map` genesis tickers and consuming modules

aljo242:
Create `AfterMarketGenesis()` hook which can be called by any module integrating w/ the market map.

This can be used to verify that the state is consistent between the modules.  In the `x/oracle` implementation, we ensure that the `x/oracle` state is a superset of the `x/marketmap` ticker state.


tests/simapp/config.go
```diff
@@ -174,6 +174,7 @@ var (
 						oracletypes.ModuleName,
 						incentivetypes.ModuleName,
 						alerttypes.ModuleName,
+						// market map genesis must be called AFTER all consuming modules (i.e. x/oracle, etc.)
```
davidterpay, line 177: can you give a go doc on why?

davidterpay (approved):

PR #192, opened 2024-02-27
fix: Make `VerifyVoteExtension` stateless [BLO-945]

g3ndo9kar9:
## In This PR
- I remove stateful validation from `VerifyVoteExtension`, this is necessary as votes are extended against block `h` (we optimistically apply proposals in `ExtendVote`), while we are only able to verify against `h - 1` state.

### DIFF
--- a/abci/ve/utils.go
+++ b/abci/ve/utils.go
@@ -14,26 +14,14 @@ import (
 
 // ValidateOracleVoteExtension validates the vote extension provided by a validator.
 func ValidateOracleVoteExtension(
-	ctx sdk.Context,
 	ve vetypes.OracleVoteExtension,
 	strategy currencypair.CurrencyPairStrategy,
 ) error {
 	// Verify prices are valid.
-	for id, bz := range ve.Prices {
+	for _, bz := range ve.Prices {
 		// Ensure that the price bytes are not too long.
 		if len(bz) > slinkyabci.MaximumPriceSize {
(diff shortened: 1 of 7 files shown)


abci/ve/utils.go
```diff
@@ -7,34 +7,20 @@ import (
...
 	// Verify prices are valid.
-	for id, bz := range ve.Prices {
+	for _, bz := range ve.Prices {
 		// Ensure that the price bytes are not too long.
 		if len(bz) > slinkyabci.MaximumPriceSize {
 			return fmt.Errorf("price bytes are too long: %d", len(bz))
 		}
-
-		// Ensure that the currency pair ID is valid.
-		cp, err := strategy.FromID(ctx, id)
```
Eric-Warehime, line 29: Seems like now we definitely don't want to do this TODO https://github.com/skip-mev/slinky/blob/2144d4ffcf660650fcaa13b9665afc8cfee456b5/abci/strategies/aggregator/vote_aggregator.go#L182

Eric-Warehime, line 29: Only concern I have with removing this is whether or not it makes the following possible:

The size of VEs for a given block can now be made arbitrarily large by any validator including a bunch of random currency pairs in their VE. Without this check those invalid price pairs are still safely ignored, but are committed to chain as part of the block which now has a bunch arbitrarily large data in its VE (potentially so large that it DOS's other txs in the mempool?

PR #197, opened 2024-02-28
feat: attach errors to provider metrics

aljo242:
Created new types `UnresolvedResult`, `ErrorWithCode` for easy propagation of error codes to the `base` provider level


providers/base/metrics/provider.go
```diff
@@ -19,6 +19,8 @@ const (
 	ProviderTypeLabel = "type"
 	// StatusLabel is a label for the status of a provider response.
 	StatusLabel = "status"
+	// ErrorLabel is a label for and error of a failed provider response.
+	ErrorLabel = "error"
```
davidterpay: I would generally recommend not logging direct errors from the providers since these may vary significantly across the board. We should set up our instrumentation to notify us anytime the zapper logger does get an error in the response tho. I would talk to @Zygimantass about this.

aljo242: I agree w this.  Will push changes where we can standardize with error codes

config/local/market.json
```diff
@@ -694,16 +694,16 @@
     "ADA/USD": {
       "providers": [
         {
-          "name": "coinbase_websocket",
-          "off_chain_ticker": "ADA-USD"
+          "name": "kraken",
```
davidterpay: could you revert this so that there isnt a diff

providers/types/errors.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrorCode is a type alias for an int error code.
+type ErrorCode int
+
+const (
+	OK                         ErrorCode = 0
```
davidterpay, line 12: chad

davidterpay: nit: I would split up the errors into the directories that compose specific errors. for example i think it makes more sense for the API errors to be in the base/api directory, ws errors in the base/ws directory, etc.

providers/types/response.go
```diff
@@ -18,26 +18,33 @@ type ResponseValue interface {
...
-// Result is the result of a single requested ID.
-type Result[V ResponseValue] struct {
+// ResolvedResult is the result of a single requested ID.
+type ResolvedResult[V ResponseValue] struct {
 	// Value is the value of the requested ID.
 	Value V
 	// Timestamp is the timestamp of the value.
 	Timestamp time.Time
+	// Code is the result code from the request.
+	Code ErrorCode
```
davidterpay, line 32: this is mega-chad

providers/types/errors.go
```diff
@@ -0,0 +1,108 @@
...
+	ErrorPongFailed            ErrorCode = 7
+	ErrorInvalidRequest        ErrorCode = 8
+	ErrorInvalidArgument       ErrorCode = 9
+	ErrorUnableToCreateURL     ErrorCode = 10
+	ErrorWebsocketStartFail    ErrorCode = 11
+	ErrorInvalidAPIChains      ErrorCode = 12
+	ErrorNoResponse            ErrorCode = 13
+	ErrorInvalidResponse       ErrorCode = 14
+	ErrorInvalidChainID        ErrorCode = 15
+	ErrorFailedToParsePrice    ErrorCode = 16
```
davidterpay: Does it make sense for this to just be `ErrorFailedToParse`

davidterpay: actually ignore, fine with keeping price specific stuff here.

providers/base/api/handlers/api_query_handler.go
```diff
@@ -198,17 +204,37 @@ func (h *APIQueryHandlerImpl[K, V]) subTask(
...
+			if resp != nil {
+				status = providertypes.ErrorCode(resp.StatusCode)
+			}
+
+			h.writeResponse(responseCh, providertypes.NewGetResponseWithErr[K, V](ids,
+				providertypes.NewErrorWithCode(
+					errors.ErrDoRequestWithErr(err),
+					status,
+				)),
+			)
```
davidterpay: a bit of a footgun here is that we return a separate codes for http rate limits and the provider error code. Im pretty sure the request handler will return an error anytime it doesnt get a 200, so its likely we never even reach the code below. tldr, i think just having the http status code included in sufficient.

davidterpay: not critical since we can change this but im curious to hear your thoughts.

aljo242: the error returned from the API handler will include the code right?

providers/apis/coinbase/api_handler.go
```diff
@@ -84,26 +86,34 @@ func (h *APIHandler) ParseResponse(
...
+			providertypes.NewErrorWithCode(fmt.Errorf("unknown ticker %s", ticker.String()), providertypes.ErrorUnknownPair),
+		)
 	}
 
 	// Parse the response into a CoinBaseResponse.
 	var result CoinBaseResponse
 	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
-		return types.NewPriceResponseWithErr(tickers, err)
+		return types.NewPriceResponseWithErr(tickers,
+			providertypes.NewErrorWithCode(err, providertypes.ErrorInvalidResponse),
```
davidterpay: Should this be a failed to decode error?

providers/apis/marketmap/api_handler.go
```diff
@@ -59,28 +61,53 @@ func (h *APIHandler) ParseResponse(
...
 	}
 
 	// Parse the response body into a market map object.
 	var market mmtypes.GetMarketMapResponse
 	if err := json.NewDecoder(resp.Body).Decode(&market); err != nil {
-		return types.NewMarketMapResponseWithErr(chains, fmt.Errorf("failed to parse market map response: %w", err))
+		return types.NewMarketMapResponseWithErr(chains,
+			providertypes.NewErrorWithCode(
+				fmt.Errorf("failed to parse market map response: %w", err),
+				providertypes.ErrorInvalidResponse,
```
davidterpay: Failed to Decode

providers/static/api_handler.go
```diff
@@ -71,7 +73,10 @@ func (s *MockAPIHandler) ParseResponse(
 		if price, ok := s.exchangeRates[ticker]; ok {
 			resolved[ticker] = types.NewPriceResult(price, time.Now())
 		} else {
-			unresolved[ticker] = fmt.Errorf("failed to resolve ticker %s", ticker)
+			err := fmt.Errorf("failed to resolve ticker %s", ticker)
+			unresolved[ticker] = providertypes.UnresolvedResult{
+				ErrorWithCode: providertypes.NewErrorWithCode(err, providertypes.ErrorUnknown),
```
davidterpay: ErrorUnknownPair?

davidterpay (approved):

davidterpay (approved):
small nits

PR #198, opened 2024-02-28
fix: add sleep function to API task execution

aljo242:


providers/base/api/handlers/api_query_handler.go
```diff
@@ -7,16 +7,19 @@ import (
...
 
 	"github.com/skip-mev/slinky/oracle/config"
 	"github.com/skip-mev/slinky/providers/base/api/errors"
 	"github.com/skip-mev/slinky/providers/base/api/metrics"
 	providertypes "github.com/skip-mev/slinky/providers/types"
 )
 
+const (
+	rateLimitMultiplier = 10
+)
```
davidterpay, line 21: what is this constant doing?

aljo242: its just splitting an extra multipliet so the eqn becomes

sleeptime := refresh / (num req * rateLimitMultiplier)

to ensure we are always significantly shorter than the refresh rate for the API handler in general

aljo242:
Closing for now

PR #202, opened 2024-03-04
feat: Orchestrator with live updating

davidterpay:
closes BLO-955


g3ndo9kar9:
Can u change the name of this [file](https://github.com/skip-mev/slinky/blob/main/oracle/orchestrator/orachestrator_test.go)

providers/base/fetch.go
```diff
@@ -126,21 +129,27 @@ func (p *Provider[K, V]) startWebSocket(ctx context.Context, subIDs []K, respons
...
+				p.logger.Info("web socket stopped via context")
 				return ctx.Err()
 			default:
+				if restarts > 0 {
+					p.logger.Debug("restarting websocket query handler", zap.Int("num_restarts", restarts))
+
+					// If the websocket query handler returns, then the connection was closed. Wait for
+					// a bit before trying to reconnect.
+					time.Sleep(p.wsCfg.ReconnectionTimeout)
+				}
```
davidterpay: This should speed up shutdown time rn which otherwise would trigger the reconnection sleep.

providers/base/provider.go
```diff
@@ -198,11 +197,18 @@ func (p *Provider[K, V]) GetData() map[K]providertypes.Result[V] {
 // Type returns the type of data handler the provider uses.
 func (p *Provider[K, V]) Type() providertypes.ProviderType {
 	switch {
-	case p.apiCfg.Enabled && p.api != nil:
+	case p.apiCfg.Enabled:
```
davidterpay, line 216: removing the p.api so that this function is thread-safe.

providers/base/utils.go
```diff
@@ -0,0 +1,24 @@
+package base
+
+import (
+	"fmt"
+
+	providertypes "github.com/skip-mev/slinky/providers/types"
+)
+
+// createResponseCh creates the response channel for the provider.
+func (p *Provider[K, V]) createResponseCh() error {
```
davidterpay, line 10: eventually, we can extend this to have a thread safe abstracted channel that we can verifiably know if it is closed or full before writes happen.

providers/base/updater.go
```diff
@@ -1,91 +0,0 @@
-package base
```
davidterpay, line 1: Moving all of this logic to the base provider since this abstraction is probablly too much.

davidterpay:
I am going to split this up into two PRs

PR #208, opened 2024-03-06
feat: Add optional modifiable delimiter to CurrencyPair

Eric-Warehime:
Makes the delimiter a property of the CurrencyPair type and configurable in order to support chains that want (or already have) their quotes structured via `A-B` or `A/B` etc.


pkg/types/currency_pair.go
```diff
@@ -6,17 +6,24 @@ import (
...
 }
 
+func (cp CurrencyPair) WithDelimiter(delimiter string) CurrencyPair {
+	cp.Delimiter = delimiter
+	return cp
+}
+
 // ValidateBasic checks that the Base / Quote strings in the CurrencyPair are formatted correctly, i.e.
 // Base + Quote are non-empty, and are in upper-case.
 func (cp *CurrencyPair) ValidateBasic() error {
```
aljo242, line 29: Is there a set of valid characters we should allow as delimiters?

aljo242: we should also limit it to be a string of `len == 1`

Eric-Warehime: Yeah right now I'm just considering the canonical implementation as `/` but obviously some people use `-` or I guess you could use `_` or conceivable anything else that's not also expected to be included in a base or quote.

PR #215, opened 2024-03-06
feat: Prune invalid VoteExtensions in PrepareProposal [BLO-944]

g3ndo9kar9:
Closes BLO-993

## In This PR
- I prune all invalid `VoteExtensions` from the `LocalLastCommit` before constructing a proposal
- This protects ourself from [this](https://github.com/cometbft/cometbft/issues/2523) comet bug


abci/proposals/validate.go
```diff
@@ -1,12 +1,82 @@
...
+
+	// tally total voting-power of valid votes
+	var (
+		votingPowerInCommit, totalVotingPower uint64
+	)
+
+	// Validate all oracle vote extensions.
+	for i, vote := range extendedCommitInfo.Votes {
+		// sum the total voting power
+		totalVotingPower += uint64(vote.Validator.Power)
```
Eric-Warehime, line 39: This voting power is now validated in upstream cosmos-sdk correct? So it's "trustable" in that sense?

abci/proposals/validate.go
```diff
@@ -1,12 +1,82 @@
...
+		// validate the vote-extension
+		if err := validateVoteExtension(verifyVECtx, vote, h.voteExtensionCodec, h.currencyPairStrategy); err != nil {
+			h.logger.Error(
+				"failed to validate vote extension",
+				"err", err,
+				"validator", vote.Validator.Address,
+			)
+
+			// failed to validate this vote-extension, mark it as absent in the original commit
+			vote.BlockIdFlag = cometproto.BlockIDFlagAbsent
```
Eric-Warehime, line 54: Is this the correct behavior? Based on my understanding, this will change how the x/sla module reasons about validators' behavior w/r/t their commitments. 

Does setting absent here mean a validator will be counted as having voted in consensus, but not submitted a VE i.e. a violation of the SLA for that round?

abci/proposals/validate.go
```diff
@@ -1,12 +1,82 @@
...
+			vote.VoteExtension = []byte{}
+			extendedCommitInfo.Votes[i] = vote
+		} else {
+			// tally the voting power of this vote
+			votingPowerInCommit += uint64(vote.Validator.Power)
+		}
+	}
+
+	// ensure that the valid vote-extensions compose a super-majority of voting-power
+	if requiredVP := (totalVotingPower * 2 / 3) + 1; votingPowerInCommit < requiredVP {
```
Eric-Warehime, line 65: Do we have this same calculation in another place? If so can we consolidate?

abci/proposals/validate.go
g3ndo9kar9: No this indicates that the proposer did not receive this validator's vote

abci/proposals/validate.go
g3ndo9kar9: yes

PR #217, opened 2024-03-06
feat: add `update` tx to marketmap

aljo242:
Closes BLO-970


proto/slinky/marketmap/v1/tx.proto
```diff
@@ -36,11 +39,26 @@ message CreateMarket {
...
+  Ticker ticker = 1 [ (gogoproto.nullable) = false ];
+
+  // Providers maps provider names to their off-chain
+  // representations for the given ticker of the message.
+  Providers providers = 2 [ (gogoproto.nullable) = false ];
+
+  // Paths is the list of convertable markets that will be used to convert the
+  // prices of a set of tickers to a common ticker.
+  Paths paths = 3 [ (gogoproto.nullable) = false ];
+}
```
davidterpay: Since this has direct overlap with `CreateMarket` does it make sense to just have a single `Market` struct?

x/marketmap/keeper/keeper.go
```diff
@@ -251,5 +314,30 @@ func (k *Keeper) ValidateState(ctx sdk.Context, creates []types.CreateMarket) er
...
+				if err != nil {
+					return err
+				}
+
+				if !has {
+					return fmt.Errorf("currency pair %s in path %s does not exist", cp.String(), path.ShowRoute())
+				}
+			}
+		}
+	}
```
davidterpay: This is more or less copied no? I think it makes more sense to just define a single `Market`struct so that we dont have dup code.

x/marketmap/types/msg.go
```diff
@@ -53,3 +54,46 @@ func (m *MsgUpdateMarketMap) ValidateBasic() error {
...
+			// check for duplicate providers
+			if _, seen := seenProviders[provider.Name]; seen {
+				return fmt.Errorf("duplicate provider found: %s", provider.Name)
+			}
+			seenProviders[provider.Name] = struct{}{}
+		}
+	}
+
+	return nil
+}
```
davidterpay: same as other comment.

davidterpay (commented):
left some comments

PR #218, opened 2024-03-06
feat: Synchronization of the base provider resources [BLO-975]

davidterpay:
Basic idea of this PR is that we can better synchronize all of the go routines and channels by referencing the existing context attached to a provider. This effectively makes it such that
- responses will not be written to if the context is expired or canceled.
- we can have better thread management by using wait groups associated with the two main routines in the provider
- providers us an easier way of reasoning about the lifecycle of the provider if everything is associated with a context.


providers/base/provider.go
```diff
@@ -53,24 +51,28 @@ type Provider[K providertypes.ResponseKey, V providertypes.ResponseValue] struct
 	// metrics is the metrics implementation for the provider.
 	metrics providermetrics.ProviderMetrics
 
-	// updater is the config updater that is used to fetch the configuration for the provider.
-	updater ConfigUpdater[K, V]
+	// fetchCtx is the context the context for the fetch function.
```
Eric-Warehime: nit: 
```suggestion
	// fetchCtx is the context for the fetch function.
```

providers/base/api/handlers/api_query_handler.go
```diff
@@ -186,19 +190,19 @@ func (h *APIQueryHandlerImpl[K, V]) subTask(
...
-			h.writeResponse(responseCh, providertypes.NewGetResponseWithErr[K, V](ids, errors.ErrCreateURLWithErr(err)))
+			h.writeResponse(ctx, responseCh, providertypes.NewGetResponseWithErr[K, V](ids, errors.ErrCreateURLWithErr(err)))
 			return nil
 		}
 
 		h.logger.Debug("created url", zap.String("url", url))
 
 		// Make the request.
-		ctx, cancel := context.WithTimeout(ctx, h.config.Timeout)
+		fetchCtx, cancel := context.WithTimeout(ctx, h.config.Timeout)
```
Eric-Warehime: The naming of this is a little bit confusing with the main `Provider.fetchCtx`

providers/base/fetch.go
```diff
@@ -15,57 +15,44 @@ import (
...
 
 // startAPI is the main loop for the provider. It is responsible for fetching data from the API
 // and updating the data.
-func (p *Provider[K, V]) startAPI(ctx context.Context, responseCh chan<- providertypes.GetResponse[K, V]) error {
+func (p *Provider[K, V]) startAPI(ctx context.Context) error {
 	p.logger.Info("starting api query handler")
 
 	// Start the data update loop.
+	handler := p.GetAPIHandler()
+	ids := p.GetIDs()
```
Eric-Warehime, line 41: It seems like the problem you're running into is keeping state in sync between the `IDs` in the base provider and the `ProviderMarketmap` in the data handler.

My sense is that all of this would be much easier if you just moved all state (i.e. the `ProviderMarketMap`) to the base provider and made any `APIDataHandler` functionality stateless. 

I think that would completely eliminate the problem of having to keep state in sync between fetch and main provider.

Overall though using contexts to manage stop/start is much cleaner than the message handler channel based stuff we had before.

Eric-Warehime (approved):

PR #219, opened 2024-03-07
feat: GeckoTerminal Provider MOG

davidterpay:


oracle/constants/tickers.go
```diff
@@ -29,6 +29,7 @@ var (
 	LIDO_USD      = mmtypes.NewTicker("LDO", "USD", 8, 1)
 	LITECOIN_USD  = mmtypes.NewTicker("LTC", "USD", 8, 1)
 	MAKER_USD     = mmtypes.NewTicker("MKR", "USD", 8, 1)
+	MOG_USD       = mmtypes.NewTicker("MOG", "USD", 18, 1)
```
aljo242, line 32: we are so back

davidterpay: we are so mogging

aljo242 (approved):

PR #228, opened 2024-03-12
fix: market map validation

aljo242:
Closes BLO-984
Closes BLO-972


pkg/types/currency_pair.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 const (
-	ethereum = "ETHEREUM"
+	ethereum         = "ETHEREUM"
+	MaxCPFieldLength = 128
```
aljo242, line 10: arbitrary lengths

x/marketmap/types/ticker.go
```diff
@@ -17,6 +17,8 @@ const (
 	DefaultMinProviderCount = 1
 	// MaxPathLength is the maximum length of a path for a ticker conversion.
 	MaxPathLength = 3
+	// MaxMetadataJSONFieldLength is the maximum length of the MetadataJSON field.
+	MaxMetadataJSONFieldLength = 4096
```
davidterpay: Are we confident this is sufficient for arbitrary json?

davidterpay (approved):

x/marketmap/types/ticker.go
aljo242: no lol

PR #230, opened 2024-03-12
feat: dYdX Price Aggregation

davidterpay:
closes BLO-964

Please see the updated read-me for a overview of how this new aggregation works.


aggregator/aggregator.go
```diff
@@ -27,7 +27,7 @@ type (
 // it is assumed to be called concurrently in data fetching goroutines. The DataAggregator
 // requires one of either an aggregateFn or aggregateFnFromContext to be set.
 type DataAggregator[K comparable, V any] struct {
-	mtx sync.RWMutex
+	mtx sync.Mutex
```
aljo242, line 30: Context on this change?

pkg/math/oracle/README.md
```diff
@@ -1,21 +1,220 @@
-# Oracle Price Conversion Aggregation
+# Oracle Price Aggregator
 
 ## Overview
 
-The aggregation function defined in this package is used to convert a set of price feeds into a common set of prices. This is done by converting all prices to a common decimal precision and then aggregating the prices based on the price feed conversions defined for each price feed. Note that only tickers that provide an optional set of Paths will be considered for the following conversion aggregation. If a ticker does not have a set of paths, the final price outputted will be the median of the price feeds.
+The aggregator defined in this package is used to convert a set of price feeds into a common set of prices. The prices used in aggregation contain
```
aljo242, line 5: Do we want to keep the old aggregation function as well?  Leave both as options for consumers?

pkg/math/oracle/README.md
```diff
@@ -1,21 +1,220 @@
...
+* USDT/USD
+    * COINBASE USDT/USD
+    * COINBASE USDC/USDT ^ -1
+    * BINANCE USDT/USD
+    * KUCOIN BTC/USDT ^-1 * INDEX BTC/USD
+
+A few important considerations:
+
+1. Each ticker (BTC/USD, ETH/USD, USDT/USD) can have a configured `MinimumProviderCount` which is the minimum number of providers that are required to calculate the price of the ticker.
+2. Each path must that is not a direct conversion (e.g. BTC/USD) must configure the second operation to utilize the `index` price i.e. [`IndexPrice`](./median.go).
```
aljo242:
```suggestion
2. Each path that is not a direct conversion (e.g. BTC/USD) must configure the second operation to utilize the `index` price i.e. [`IndexPrice`](./median.go).
```

pkg/math/oracle/median.go
```diff
@@ -33,279 +41,170 @@ func NewMedianAggregator(logger *zap.Logger, cfg mmtypes.MarketMap) (*MedianAggr
...
+func (m *MedianAggregator) AggregatedData() {
+	updatedPrices := make(types.TickerPrices)
+	for ticker, paths := range m.cfg.Paths {
+		target, ok := m.cfg.Tickers[ticker]
 		if !ok {
 			m.logger.Error(
-				"failed to get ticker",
-				zap.String("ticker", tickerStr),
+				"failed to get ticker; skipping aggregation",
+				zap.String("ticker", ticker),
```
aljo242, line 69: Let's cover this case

pkg/math/oracle/median.go
```diff
@@ -33,279 +41,170 @@ func NewMedianAggregator(logger *zap.Logger, cfg mmtypes.MarketMap) (*MedianAggr
...
+	if len(paths.Paths) == 0 {
+		m.logger.Error(
+			"no conversion paths",
+			zap.String("ticker", target.String()),
+		)
 
+		return nil
+	}
+
+	convertedPrices := make([]*big.Int, 0)
```
```suggestion
	convertedPrices := make([]*big.Int, 0, len(paths.Paths))
```

set capacity but not Len

pkg/math/oracle/median.go
```diff
@@ -33,279 +41,170 @@ func NewMedianAggregator(logger *zap.Logger, cfg mmtypes.MarketMap) (*MedianAggr
...
+// to adjust the price by the index price of the asset. If the index price is not available, we
+// return an error.
+func (m *MedianAggregator) CalculateAdjustedPrice(
 	target mmtypes.Ticker,
-	path mmtypes.Path,
-	medians types.TickerPrices,
+	operations []mmtypes.Operation,
 ) (*big.Int, error) {
-	if err := path.ValidateBasic(); err != nil {
+	price, err := m.GetProviderPrice(operations[0])
```
aljo242, line 180: We should note that this function assumes that `operations` is NOT nil or add a check

pkg/math/oracle/median.go
```diff
@@ -33,279 +41,170 @@ func NewMedianAggregator(logger *zap.Logger, cfg mmtypes.MarketMap) (*MedianAggr
...
+	if err != nil {
 		return nil, err
 	}
 
-	// Ensure that the conversion is valid.
-	if !path.Match(target.String()) {
-		return nil, fmt.Errorf("path does not match target %s: %s", target.String(), path.String())
+	// If we have a single operation, then we can simply return the price. This implies that
+	// we have a direct conversion from the base ticker to the target ticker.
+	if len(operations) == 1 {
```
aljo242, line 187: we should ensure that:

```
foreach path
     len(path.operations) == 1 OR == 2

```

x/marketmap/types/market.go
```diff
@@ -30,42 +28,19 @@ func (mm *MarketMap) ValidateBasic() error {
...
-		if err != nil {
-			return err
+	// check if all providers refer to tickers
+	for tickerStr, providers := range mm.Providers {
+		// check if the ticker is supported
+		if _, ok := mm.Tickers[tickerStr]; !ok {
+			return fmt.Errorf("provider %s refers to an unsupported ticker", tickerStr)
 		}
 
-		if err := paths.ValidateBasic(cp); err != nil {
```
aljo242, line 51: What is there reason for removing most of this logic?

pkg/math/oracle/README.md
davidterpay: IMO we're really only using two across deployments atm, raw medianization and dYdX. I don't particularly want to maintain other versions that are not currently being used.

aggregator/aggregator.go
davidterpay: The aggregator doesn't have multiple concurrent reads/writes so i think using a raw mutex is preferrable. happy to change this back.

pkg/math/oracle/median.go
davidterpay: We have this check in the parent function that calls this.

x/marketmap/types/market.go
davidterpay: yea, we can have USDC/USDT included in a path for USDT/USD. I think we should move this logic into each aggregator that might use the paths field in unique ways.

pkg/math/oracle/median.go
davidterpay: This implies an invalid config, which is going to come in a follow up PR.

pkg/math/oracle/median.go
davidterpay: added

pkg/math/oracle/median.go
davidterpay: modified so that this is verified in the same func

x/marketmap/types/market.go
aljo242: I see so they cyclic stuff is all gone?

pkg/math/oracle/median.go
```diff
@@ -33,279 +41,166 @@ func NewMedianAggregator(logger *zap.Logger, cfg mmtypes.MarketMap) (*MedianAggr
...
 ) []*big.Int {
-	convertedPrices := make([]*big.Int, 0)
+	m.logger.Debug("calculating converted prices", zap.String("ticker", target.String()))
+	if len(paths.Paths) == 0 {
+		m.logger.Error(
+			"no conversion paths",
+			zap.String("ticker", target.String()),
+		)
+
+		return nil
```
aljo242, line 129: Is this an error or a valid scenario?

We are logging an error but returning `nil`.  If we want to handle this error gracefully, we should return a typed error and then have the calling function handle that.

pkg/math/oracle/median.go
```diff
@@ -8,17 +8,25 @@ import (
...
 	"github.com/skip-mev/slinky/pkg/math/median"
-	slinkytypes "github.com/skip-mev/slinky/pkg/types"
 	mmtypes "github.com/skip-mev/slinky/x/marketmap/types"
 )
 
+const (
+	// MaxConversionOperations is the maximum number of conversion operations that can be used
+	// to convert a set of prices to a common ticker. This implementation only supports a maximum
+	// of 2 conversion operations - either a direct conversion or a conversion using the index price.
+	MaxConversionOperations = 2
```
aljo242, line 18: just noting we need to enforce this limit in `x/marketmap` as well

aljo242 (approved):

PR #233, opened 2024-03-13
feat: constrained sw-median vote weighting

Eric-Warehime:
Adds algorithm which is complementary to our SW-median w/ threshold. 

It introduces weighting of validator power based on difference in ppm the vote is from the pre-computed stake-weighted median and additionally how far the SWM is from the previous price.

Users of this algorithm can provide their own "ppm" via wrapping our Threshold function and providing a custom ppm parameter based on the currency pair in the PriceInfo.

Also refactored extraction of `PriceInfo` since it is re-used for this implementation.

BLO-982


pkg/math/voteweighted/threshold.go
```diff
@@ -0,0 +1,87 @@
...
+
+// SignsDiffer determines whether newPrice and newPrime move in different directions relative to old.
+func SignsDiffer(old *big.Int, newPrice *big.Int, newPrime *big.Int) bool {
+	newSign := (&big.Int{}).Sub(old, newPrice).Sign()
+	newPrimeSign := (&big.Int{}).Sub(old, newPrime).Sign()
+	return newSign != newPrimeSign
+}
+
+// CalculateRelativeTicks computes the ticks of newPrime relative to the tick size of newPrice.
+func CalculateRelativeTicks(old *big.Int, newPrice *big.Int, newPrime *big.Int, ppm *big.Int) *big.Int {
```
g3ndo9kar9, line 23: nit: can we name these oldPrice, aggregatedPrice, validatorPrice (or something similar) to indicate what values these variables correspond to intuitively

pkg/math/voteweighted/threshold.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+var (
+	Zero       = big.NewInt(0)
+	Two        = big.NewInt(2)
+	OneMillion = big.NewInt(1_000_000)
+)
+
+// SignsDiffer determines whether newPrice and newPrime move in different directions relative to old.
+func SignsDiffer(old *big.Int, newPrice *big.Int, newPrime *big.Int) bool {
```
g3ndo9kar9, line 16: nit: can we name these oldPrice, aggregatedPrice, validatorPrice (or something similar) to indicate what values these variables correspond to intuitively

pkg/math/voteweighted/voteweighted.go
```diff
@@ -194,3 +209,63 @@ func ComputeMedian(priceInfo PriceInfo) *big.Int {
...
+					logger.Error(
+						"found currency pair in votes which doesn't exist in module state",
+						"currency pair", currencyPair.String(),
+						"error", err,
+					)
+					continue
+				}
+				currentPrice := quote.Price.BigInt()
+				// newPrice is the calculated stake-weighted median
+				newPrice := ComputeMedian(info)
```
g3ndo9kar9, line 248: In the event that we don't have enough stake  to compute a stake weighted median (i.e info.TotalWeight < 2/3 stake weight report vote-extensions with prices), should we even compute a stake weighted median / perform the threshold calculation?

pkg/math/voteweighted/voteweighted.go
```diff
@@ -160,6 +105,76 @@ func Median(
 	}
 }
 
+func PriceInfoFromProviders(
```
g3ndo9kar9, line 108: nit: this should be `PriceInfoFromValidatorStake`, or similar, not sure why the notion of a `Provider` is involved here

pkg/math/voteweighted/threshold.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// ThresholdWeightCalc computes the amount of weight taken into account for the given price update.
+// It follows an algorithm which requires increasing vote price correlation as the percent change in
+// the price increases.
+func ThresholdWeightCalc(
+	currentPrice *big.Int,
+	proposedPrice *big.Int,
+	ppm *big.Int,
+	priceInfo PriceInfo,
```
g3ndo9kar9, line 65: Does priceInfo have the current price for all validators who did not post a price for the considered ticker? Otherwise , the (zero, min_price_change_ppm) price-change assumption for validators who did not vote is not considered in our algorithm?

pkg/math/voteweighted/threshold.go
```diff
@@ -0,0 +1,87 @@
...
+// CalculateTicks computes the ticks of newPrice relative to old.
+func CalculateTicks(old *big.Int, newPrice *big.Int, ppm *big.Int) *big.Int {
+	/*
+		1_000_000 * abs(old-newPrice)
+		________________________
+			   (old*ppm)
+	*/
+
+	// abs(old-newPrice)
+	delta := (&big.Int{}).Abs((&big.Int{}).Sub(old, newPrice))
```
aljo242, line 49:
```suggestion
	delta := (new(big.Int)).Abs((&big.Int{}).Sub(old, newPrice))
```

nit:  Sometimes we use `new(big.Int)` and sometimes we use (&big.Int{}).

Let's just unify

pkg/math/voteweighted/voteweighted.go
```diff
@@ -63,67 +67,8 @@ func Median(
...
-					}),
-					TotalWeight: cpInfo.TotalWeight.Add(voteWeight),
-				}
-			}
-		}
+		// Calculate map of CurrencyPair to PriceInfo
+		priceInfo := PriceInfoFromProviders(ctx, logger, validatorStore, providers)
 
 		// Iterate through all prices and compute the median price for each asset.
 		prices := make(map[slinkytypes.CurrencyPair]*big.Int)
```
aljo242, line 74:
```suggestion
		prices := make(map[slinkytypes.CurrencyPair]*big.Int, len(priceInfo))
```

init with capacity

pkg/math/voteweighted/voteweighted.go
```diff
@@ -194,3 +209,63 @@ func ComputeMedian(priceInfo PriceInfo) *big.Int {
...
+	thresholdDetermination ThresholdDetermination,
+) aggregator.AggregateFnFromContext[string, map[slinkytypes.CurrencyPair]*big.Int] {
+	return func(ctx sdk.Context) aggregator.AggregateFn[string, map[slinkytypes.CurrencyPair]*big.Int] {
+		// providers is a map of consensus-address to map of CurrencyPair to Price
+		// it contains the mappings of price points for each validator
+		return func(providers aggregator.AggregatedProviderData[string, map[slinkytypes.CurrencyPair]*big.Int]) map[slinkytypes.CurrencyPair]*big.Int {
+			// Calculate map of CurrencyPair to PriceInfo
+			priceInfo := PriceInfoFromProviders(ctx, logger, validatorStore, providers)
+			// Iterate through all prices and compute the median price for each asset.
+			prices := make(map[slinkytypes.CurrencyPair]*big.Int)
```
aljo242, line 227:
```suggestion
			prices := make(map[slinkytypes.CurrencyPair]*big.Int, len(priceInfo))
```

init with capacity

pkg/math/voteweighted/threshold.go
BrendanChou: nit: `newPrime` and `newPrice` are also a bit hard for eyes to distinguish quickly, leading to some fatigue

pkg/math/voteweighted/threshold.go
```diff
@@ -0,0 +1,87 @@
...
+	"cosmossdk.io/math"
+)
+
+var (
+	Zero       = big.NewInt(0)
+	Two        = big.NewInt(2)
+	OneMillion = big.NewInt(1_000_000)
+)
+
+// SignsDiffer determines whether newPrice and newPrime move in different directions relative to old.
```
BrendanChou, line 15: nit: you might have to explain whether this function returns true or false if one or both of the values is zero. It seems to do this weird thing where if exactly one is zero, it returns true, but false if both are zero

PR #234, opened 2024-03-13
feat: MarketMap validation for oracle aggregation function

davidterpay:
closes BLO-1002


pkg/math/oracle/utils.go
```diff
@@ -55,3 +56,81 @@ func (m *MedianAggregator) GetProviderPrice(
...
+
+// ValidateMarketMap validates the market map configuration and its expected configuration for
+// this aggregator. In particular, this will
+//
+//  1. Ensure that the market map is valid (ValidateBasic). This ensure's that each of the provider's
+//     markets are supported by the market map.
+//  2. Ensure that each path has a corresponding ticker.
+//  3. Ensure that each path has a valid number of operations.
+//  4. Ensure that each operation has a valid ticker and that the provider supports the ticker.
+func ValidateMarketMap(
```
aljo242: Since this is the only aggregator we are going to support AND this function does not rely on anything aggregator-specific AFAIK, why not just include this in mm.ValidateBasic()?

aljo242: separating out only makes sense to me if the plan is to support multiple aggregation strategies

davidterpay: I prefer deferring whether we want this to be the canonical one to a later time, I can see us porting over the logic in the future.

aljo242: Makes sense - holding over in case we wanna factor out.  I still think this should live in `x/marketmap/types` as a function called inside of `ValidateBasic`.  In the event that we want to move it elsewhere, it will be easy, but as things currently are, this encompasses the full validation for what can be on and off chain

x/marketmap/types/market.go
```diff
@@ -2,6 +2,19 @@ package types
 
 import (
 	"fmt"
+
+	pkgtypes "github.com/skip-mev/slinky/pkg/types"
```
aljo242: nit: everywhere else in the codebase we use `slinkytypes`

x/marketmap/types/market.go
```diff
@@ -40,15 +53,95 @@ func (mm *MarketMap) ValidateBasic() error {
 		}
 	}
 
-	return nil
+	switch mm.AggregationType {
+	case AggregationType_INDEX_PRICE_AGGREGATION:
+		return ValidateIndexPriceAggregation(*mm)
+	default:
+		return nil
```
aljo242, line 60: should this be an "error unsupported"?

aljo242 (approved):

x/marketmap/types/market.go
davidterpay: I left it as this for now just for simplicity, we can probably add other conditions later. Otherwise we break every other test case using the market map

PR #236, opened 2024-03-14
fix: prune invalid VEs

aljo242:
Closes BLO-1006


abci/proposals/validate.go
```diff
@@ -28,33 +32,80 @@ func (h *ProposalHandler) ValidateExtendedCommitInfo(
...
+	return extendedCommitInfo, nil
+}
+
+func validateVoteExtension(
+	ctx sdk.Context,
+	vote cometabci.ExtendedVoteInfo,
+	voteExtensionCodec codec.VoteExtensionCodec,
+	currencyPairStrategy currencypair.CurrencyPairStrategy,
+) error {
+	// vote is not voted for if VE is nil
```
g3ndo9kar9, line 94: wym

g3ndo9kar9: why not j skip non-commit votes above?

abci/proposals/validate.go
```diff
@@ -28,33 +32,80 @@ func (h *ProposalHandler) ValidateExtendedCommitInfo(
...
+			// failed to validate this vote-extension, mark it as absent in the original commit
+			vote.BlockIdFlag = cometproto.BlockIDFlagAbsent
+			vote.ExtensionSignature = nil
+			vote.VoteExtension = nil
+			extendedCommitInfo.Votes[i] = vote
 		}
 	}
 
+	// validate after pruning
+	if err := h.validateVoteExtensionsFn(ctx, extendedCommitInfo); err != nil {
```
g3ndo9kar9, line 76: It's not clear to the user whether or not this method always checks that the votes are a quorum, no? Why not perform this check in the pruning fn?

aljo242: what do you mean? This is in the pruning function.

We:

1. Prune invalid VEs and set them to absent + nil
2. check that _after_ pruning, is there still quorum

abci/proposals/validate.go
aljo242: you mean before this function is called?  So that if a `nil` VE is passed into this function it will be considered invalid?

aljo242: my thinking is that a `nil` vote IS valid as long as there is no signature.  It is valid behavior to have this kind of vote, so to me it makes sense to live in here no?

abci/proposals/validate.go
g3ndo9kar9: Ya, I'm asking why we're not explicitly checking that there's a quorum here. Instead it looks like we expect that the `ValidateVoteExtensionsFn` does this for us?

abci/proposals/validate.go
g3ndo9kar9: kk in that case, let's adjust this check to check that there shld be extensions for commit votes, and shld not for non-commit votes? Also the godoc on L95 is a bit unclear

abci/proposals/validate.go
g3ndo9kar9: Actually, this is fine, we use `validateVoteExtensionsFn` in process anyway, so this check makes sense

abci/proposals/validate.go
```diff
@@ -28,33 +32,80 @@ func (h *ProposalHandler) ValidateExtendedCommitInfo(
...
+	if vote.VoteExtension == nil && vote.ExtensionSignature == nil {
+		return nil
+	}
+
+	voteExt, err := voteExtensionCodec.Decode(vote.VoteExtension)
+	if err != nil {
+		return err
+	}
+
+	// The vote extensions are from the previous block.
```
g3ndo9kar9, line 104: nit: either remove the comment, or have it explain what we're doing here?

abci/proposals/validate.go
davidterpay: small nit, it is a bit confusing read this code and seeing `validateVoteExtension` and `validateVoteExtensionsFn`

davidterpay:
I'm personally not so big of a fan of altering the extended commit info if a given vote extension is invalid as it might have unintended consequences for our other downstream modules (SLA, Alerts).

davidterpay (approved):

Eric-Warehime (approved):

g3ndo9kar9 (approved):
lgtm

PR #240, opened 2024-03-15
feat: Lifecycle Management of Providers with Orchestrator

davidterpay:


oracle/orchestrator/lifecycle.go
```diff
@@ -0,0 +1,119 @@
...
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+// CtxErrors is a map of context errors that we check for when starting the provider.
+// We only want cancel the main error group if the context is canceled or the deadline
+// is exceeded. Otherwise failures should be handled gracefully.
```
aljo242:
```suggestion
// is exceeded. Otherwise, failures should be handled gracefully.
```

oracle/orchestrator/lifecycle.go
```diff
@@ -0,0 +1,119 @@
...
+// GeneralProvider is a interface for a provider that implements the base provider.
+type GeneralProvider interface {
+	// Start starts the provider.
+	Start(ctx context.Context) error
+	// Stop stops the provider.
+	Name() string
+}
+
+// Start starts the provider orchestrator. This will initialize the provider orchestrator
+// with the relevant price providers and market mapper, and then start all of them.
```
aljo242:
```suggestion
// with the relevant price and market map providers, and then start them.
```

oracle/orchestrator/lifecycle.go
```diff
@@ -0,0 +1,119 @@
...
+	context.DeadlineExceeded: {},
+}
+
+// GeneralProvider is a interface for a provider that implements the base provider.
+type GeneralProvider interface {
+	// Start starts the provider.
+	Start(ctx context.Context) error
+	// Stop stops the provider.
+	Name() string
+}
```
aljo242: I think this interface should be defined in `base` provider dir.  

And we should also add a type assertion that all of our providers implement this interface.

oracle/orchestrator/lifecycle.go
```diff
@@ -0,0 +1,119 @@
...
+			if r := recover(); r != nil {
+				o.logger.Error("recovered from panic", zap.Error(fmt.Errorf("%v", r)))
+			}
+		}()
+
+		// If the context is canceled, or the deadline is exceeded,
+		// we want to exit the provider and trigger the error group
+		// to exit for all providers.
+		err := p.Start(ctx)
+		if _, ok := CtxErrors[err]; ok {
```
aljo242: did we need to make `CtxErrors` a global var because it is being used like this in a function closure?

I'm hesitant to use global variables anywhere in the codebase

oracle/orchestrator/lifecycle.go
davidterpay: I was thinking we can probably just migrate most of the functionality on the base provider into the provider types directory as additional interfaces. Ill make this a non-global interface for the time being.

oracle/orchestrator/lifecycle.go
davidterpay: made it non-global

aljo242 (approved):

PR #243, opened 2024-03-18
feat: Aggregation Func updating in orchestrator

davidterpay:


pkg/math/oracle/median.go
```diff
@@ -48,6 +52,10 @@ func NewMedianAggregator(logger *zap.Logger, cfg mmtypes.MarketMap) (*MedianAggr
 //
 // The index price cache contains the previously calculated median prices.
 func (m *MedianAggregator) AggregatedData() {
+	m.indexCache = m.GetAggregatedData()
+	m.providerCache = m.GetProviderData()
+
+	m.Lock()
```
davidterpay, line 58: Place a lock so that the config cannot change till we finish execution with the current one.

pkg/math/oracle/utils.go
```diff
@@ -55,3 +56,19 @@ func (m *MedianAggregator) GetProviderPrice(
...
+
+	m.cfg = marketMap
+}
+
+// GetMarketMap returns the market map for the oracle.
+func (m *MedianAggregator) GetMarketMap() mmtypes.MarketMap {
+	m.Lock()
+	defer m.Unlock()
+
+	return m.cfg
```
aljo242: This might mean that this object is copied a bunch which we probably dont want to be doing

davidterpay: gonna use a pointer ref instead

aljo242 (approved):

PR #245, opened 2024-03-18
feat(`x/oracle`): track number of CPs removed in previous block

aljo242:
- Add `BeginBlock` which rests state

track:
- num removed CPs
- number of CPs


x/oracle/keeper/grpc_query.go
```diff
@@ -129,3 +129,21 @@ func (q queryServer) GetPrices(goCtx context.Context, req *types.GetPricesReques
 		Prices: prices,
 	}, nil
 }
+
+// RemovedCPs returns the number of currency pairs removed in the previous block.
```
davidterpay: Is there a reason we need to expose this as a gRPC query? This feels un-necesssary given that our dependencies aren't on the query server but rather the keeper itself.

x/oracle/keeper/keeper.go
```diff
@@ -62,6 +62,13 @@ type Keeper struct {
 	// indexes
 	idIndex *indexes.Multi[uint64, string, types.CurrencyPairState]
 
+	// numRemoves is the number of CPs removed in the previous block.
+	numRemoves collections.Item[uint64]
+
+	// removedCPsCounter is an in-memory counter for the number of CPs removed in the previous block.
+	// It is flushed to numRemovedCPs during the EndBlocker.
+	removedCPsCounter uint64
```
davidterpay: Is there a reason why we don't just increment `numRemoves` instead of keeping a separate counter for this? PreBlock is executed before begin block so it might just be simpler to reset this in begin block and have it be accessible directly via the keeper when preblock is called in the next height.

x/oracle/keeper/abci.go
```diff
@@ -0,0 +1,36 @@
+package keeper
+
+import (
+	"context"
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// BeginBlocker is called at the beginning of every block.  It resets the count of
```
davidterpay, line 10: Can we add testing for this?

x/oracle/keeper/msg_server.go
```diff
@@ -50,6 +50,10 @@ func (m *msgServer) AddCurrencyPairs(goCtx context.Context, req *types.MsgAddCur
 			if err != nil {
 				return nil, fmt.Errorf("error creating currency pair state: %w", err)
 			}
+
```
davidterpay: Testing

x/oracle/keeper/msg_server.go
```diff
@@ -86,6 +90,9 @@ func (m *msgServer) RemoveCurrencyPairs(goCtx context.Context, req *types.MsgRem
 
 		// delete the currency pair from state
 		m.k.RemoveCurrencyPair(ctx, cp)
+		if err := m.k.IncrementRemovedCPCounter(ctx); err != nil {
```
davidterpay: Testing

davidterpay (commented):
changes lgtm, can we add testing for it?

davidterpay (approved):

x/oracle/keeper/abci.go
```diff
@@ -0,0 +1,36 @@
...
+func (k *Keeper) BeginBlocker(goCtx context.Context) error {
+	// unwrap the context
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	removes, err := k.numRemoves.Get(ctx)
+	if err != nil {
+		return err
+	}
+
+	numCPs, err := k.numCPs.Get(ctx)
```
g3ndo9kar9, line 21: So semantically, `numCPs - numRemoves` is the actual # of currency-pairs in state at any point in time? `numRemoves` points to the latest state, while `numCPs` points to the `latestState - 1`?

abci/strategies/currencypair/types.go
```diff
@@ -16,6 +16,8 @@ type OracleKeeper interface {
 	GetCurrencyPairFromID(ctx sdk.Context, id uint64) (cp slinkytypes.CurrencyPair, found bool)
 	GetIDForCurrencyPair(ctx sdk.Context, cp slinkytypes.CurrencyPair) (uint64, bool)
 	GetPriceForCurrencyPair(ctx sdk.Context, cp slinkytypes.CurrencyPair) (oracletypes.QuotePrice, error)
+	GetRemovedCPCounter(ctx sdk.Context) (uint64, error)
+	GetCPCounter(ctx sdk.Context) (uint64, error)
```
g3ndo9kar9: What do we intend `CPCounter` to be used for?

x/oracle/types/keys.go
```diff
@@ -25,4 +26,13 @@ var (
...
 	IDIndexCurrencyPairKeyPrefix = collections.NewPrefix(3)
+
+	// NumRemovesKeyPrefix is the key-prefix under which the number of removed CPs is stored.
+	NumRemovesKeyPrefix = collections.NewPrefix(4)
+
+	// NumCPsKeyPrefix is the key-prefix under which the number CPs is stored.
+	NumCPsKeyPrefix = collections.NewPrefix(5)
+
+	// NumRemovesCodec is the collections.KeyCodec value used for the numRemoves value.
+	NumRemovesCodec = codec.KeyToValueCodec[uint64](codec.NewUint64Key[uint64]())
```
g3ndo9kar9: We don't need a codec for `NumCPs`?

g3ndo9kar9 (commented):
Does this work for messages that are executed via x/gov proposals?

x/oracle/keeper/abci.go
aljo242: correct.  Do you think it would be better to rename the variables ?

x/oracle/types/keys.go
aljo242: no - its the same codec for both.

I will rename to make this clearer

abci/strategies/currencypair/types.go
aljo242: The max Amt of CP we should expect in VEs is `CPCounter`.

The max Amt of errors we should expect in VEs is `RemovedCPCounter`

aljo242:
a lot of noise from mock regeneration

g3ndo9kar9 (approved):
lgtm

PR #246, opened 2024-03-19
feat: Start-Stop providers based on the provider market config

davidterpay:
closes BLO-956


oracle/orchestrator/lifecycle.go
```diff
@@ -33,73 +24,67 @@ func (o *ProviderOrchestrator) Start(ctx context.Context) error {
...
-
-		o.logger.Info("provider orchestrator exited successfully")
 	}
 
-	return nil
+	o.wg.Wait()
+	o.logger.Info("provider orchestrator exited successfully")
 }
 
 // execProviderFn returns a function that starts the provider. This function is used
```
aljo242: nit: function no longer returns a function

oracle/orchestrator/market_mapper.go
```diff
@@ -11,53 +11,52 @@ import (
...
-	return func() error {
-		mapper := o.GetMarketMapProvider()
-		ids := mapper.GetIDs()
-		if len(ids) != 1 {
-			o.logger.Error("market map provider can only be responsible for one chain", zap.Any("ids", ids))
-			return nil
-		}
+func (o *ProviderOrchestrator) listenForMarketMapUpdates(ctx context.Context) {
+	mapper := o.GetMarketMapProvider()
+	ids := mapper.GetIDs()
```
aljo242: mmProvider

providers/base/provider.go
```diff
@@ -100,12 +100,14 @@ func NewProvider[K providertypes.ResponseKey, V providertypes.ResponseValue](opt
 // Start starts the provider's main loop. The provider will fetch the data from the handler
 // and continuously update the data. This blocks until the provider is stopped.
 func (p *Provider[K, V]) Start(ctx context.Context) error {
-	p.logger.Info("starting provider")
+	if ctx == nil {
+		p.logger.Info("context is nil; exiting")
+		return nil
```
aljo242, line 105: shouldn't this be an error?

oracle/oracle.go
```diff
@@ -181,6 +178,15 @@ func (o *OracleImpl) fetchPrices(provider types.PriceProviderI) {
 		}
 	}()
 
+	if !provider.IsRunning() {
```
Eric-Warehime, line 181: This just reads kinda weird. If the provider wasn't running wouldn't fetchPrices not be called?

Eric-Warehime: Maybe what we want here is some kind of context check or something?

oracle/orchestrator/lifecycle.go
```diff
@@ -33,73 +24,67 @@ func (o *ProviderOrchestrator) Start(ctx context.Context) error {
...
-			o.logger.Error("provider orchestrator exited with error", zap.Error(err))
-			return err
-		}
-
-		o.logger.Info("provider orchestrator exited successfully")
 	}
 
-	return nil
+	o.wg.Wait()
+	o.logger.Info("provider orchestrator exited successfully")
```
Eric-Warehime, line 66: Will the logger provide context here such as which provider this is, or will the log line just be "provider exited"? If not I think we should add some context.

aljo242 (approved):

PR #247, opened 2024-03-19
feat: Deal with Duplicates in Response, Clean Up

davidterpay:


pkg/math/oracle/median.go
```diff
@@ -32,9 +33,9 @@ func NewMedianAggregator(logger *zap.Logger, cfg mmtypes.MarketMap) (*MedianAggr
 	}
 
 	return &MedianAggregator{
-		logger:          logger,
-		cfg:             cfg,
-		PriceAggregator: types.NewPriceAggregator(),
+		logger:         logger,
+		cfg:            cfg,
+		DataAggregator: aggregator.NewDataAggregator[string, types.TickerPrices](),
```
aljo242, line 38: are we removing the price aggregator type alias?

Isn't it still just 

```go
type PriceAggregator = DataAggregator[string, TickerPrices]
```

providers/apis/dydx/README.md
```diff
@@ -1,5 +1,358 @@
 # dYdX Market Params API
 
-## Overview
+> To learn more about the price aggregation that pairs with this API as well as the structure of the side-cars market config (Market Map), please refer to the [Median Index Price Aggregator readme](../../../pkg/math/oracle/README.md).
```
aljo242:
```suggestion
> To learn more about the price aggregation that pairs with this API as well as the structure of the side-car's market config (Market Map), refer to the [Median Index Price Aggregator readme](../../../pkg/math/oracle/README.md).
```

aljo242 (approved):

pkg/math/oracle/median.go
davidterpay: i had to remove it since the oracle expects a `DataAggregator` type directly but we inherit it so now we just use the interface instead.

PR #249, opened 2024-03-19
feat: dYdX market provider clean up + multiple provider support

davidterpay:


providers/apis/dydx/utils.go
```diff
@@ -36,7 +48,26 @@ var DefaultAPIConfig = config.APIConfig{
 	Interval:         10 * time.Second,
 	ReconnectTimeout: 2000 * time.Millisecond,
 	MaxQueries:       1,
-	URL:              "localhost:1317",
+	URL:              "https://dydx-api.polkachu.com",
```
aljo242: we will need to note how validators should configure this in the future

providers/apis/dydx/utils.go
```diff
@@ -112,128 +144,190 @@ func CreateTickerFromMarket(market dydxtypes.MarketParam) (mmtypes.Ticker, error
...
-			if err != nil {
-				return mmtypes.Paths{}, mmtypes.Providers{}, err
-			}
-			second := mmtypes.Operation{
-				CurrencyPair: cp,
-				Provider:     mmtypes.IndexPrice,
-				Invert:       false,
+		// We passively ignore errors here, as we don't want to fail the entire conversion.
+		if err != nil {
+			continue
```
aljo242: log error?

providers/apis/dydx/utils.go
davidterpay: yea i need to change this back to localhost so that we dont spam polkachu lol

providers/apis/dydx/utils.go
```diff
@@ -112,128 +144,190 @@ func CreateTickerFromMarket(market dydxtypes.MarketParam) (mmtypes.Ticker, error
...
+			Operations: []mmtypes.Operation{
+				{
+					CurrencyPair: ticker.CurrencyPair,
+					Provider:     name,
+					Invert:       false,
+				},
+			},
+		}
+		paths = append(paths, path)
+	}
```
aljo242:
```suggestion
	paths := make([]mmtypes.Path, len(exchangeNames)
	for i, name := range exchangeNames {
		paths[i] = mmtypes.Path{
			Operations: []mmtypes.Operation{
				{
					CurrencyPair: ticker.CurrencyPair,
					Provider:     name,
					Invert:       false,
				},
			},
		}
	}
```

providers/apis/dydx/utils.go
```diff
@@ -112,128 +144,190 @@ func CreateTickerFromMarket(market dydxtypes.MarketParam) (mmtypes.Ticker, error
...
+	return paths
+}
 
-			operations = append(operations, first, second)
+// InvertedConversion is a conversion with an inverted price i.e. USD/BTC ^ -1 = BTC/USD.
+func InvertedConversion(
+	ticker mmtypes.Ticker,
+	exchangeNames []string,
+) []mmtypes.Path {
+	paths := make([]mmtypes.Path, 0)
```
aljo242: same thing throughout helpers

providers/apis/dydx/utils.go
davidterpay: actually this might be a more difficult since i would have to attach the logger to the API handler, I'd prefer to have this be an exportable helper function in the case that dYdX wants to use it to test things.

aljo242 (approved):

PR #252, opened 2024-03-20
chore: Add local config check to lint workflow

Eric-Warehime:
Adds a makefile target, script, and github workflow to fail CI whenever changes to the local configs haven't been committed.


Makefile
```diff
@@ -45,6 +45,9 @@ update-local-config:
 	@echo "Updating local config..."
 	@go generate ${CONFIG_DIR}
 
+diff-local-configs: update-local-config
```
aljo242: This is going to fail everytime because it runs `update-local-config`.  The output of this is nondeterministic due to map iteration

Eric-Warehime: Yes I realized that after running it a few times :(
Maybe I can sort before serialization

aljo242 (approved):

PR #254, opened 2024-03-20
[DO NOT MERGE] feat: modify the Binance provider for OracleOps take home

Zygimantass:


g3ndo9kar9 (commented):
lgtm, looks like this is erroring every 5 mins

PR #256, opened 2024-03-20
feat: Build targets

davidterpay:
99% of this PR is generating the dYdX market config


contrib/images/slinky.sidecar.Dockerfile
```diff
@@ -11,10 +11,10 @@ RUN make build
 
 FROM ubuntu:rolling
 EXPOSE 8080
-EXPOSE 8001
+EXPOSE 8002
```
davidterpay, line 14: Is there a reason this was 8001 before hand?

config/dydx/market.json
```diff
@@ -0,0 +1,10579 @@
+{
```
davidterpay, line 1: kek

contrib/images/slinky.sidecar.e2e.Dockerfile
```diff
@@ -8,16 +8,16 @@ RUN go mod download
...
 
 RUN make build
-RUN go generate ./config
+RUN make update-local-configs
 
 
 FROM ubuntu:rolling
 EXPOSE 8080
-EXPOSE 8001
+EXPOSE 8002
```
davidterpay, line 16: Same as other question

oracle/constants/paths.go
```diff
@@ -1,379 +0,0 @@
-package constants
```
davidterpay, line 1: we no longer use any of this

cmd/config/main.go
```diff
@@ -30,63 +26,95 @@ import (
...
-	// tickersOpt defines an optional list of tickers to include in the local market
-	// config file. If this is not provided, all tickers will be included.
-	tickersOpt = flag.String(
-		"tickers",
+	// nodeURL is the URL of the validator. This is required if running the oracle with a market map provider.
+	nodeURL = flag.String(
+		"node-http-url",
 		"",
-		"optional list of tickers to include in the local market config file",
+		"URL of the dYdX node. this is required if running the oracle on the dYdX chain. this is the http address of the dYdX node.",
```
aljo242: I think this string should not make any references to dYdX.  This is just the "node market map source"

cmd/config/main.go
```diff
@@ -30,63 +26,95 @@ import (
...
-	// market config file. If this is not provided, all providers will be included.
-	providersOpt = flag.String(
-		"providers",
+	// chain defines the chain that we expect the oracle to be running on. Currently only
+	// supprting dYdX.
+	chain = flag.String(
+		"chain",
 		"",
-		"optional list of providers to include in the local market config file",
+		"chain that we expect the oracle to be running on. ex dydx.",
```
aljo242: So this is not `chainID` this is just used to switch the dydx specific logic?

cmd/config/main.go
```diff
@@ -236,123 +260,115 @@ var (
...
+
+// createOracleConfig creates an oracle config given all of the local provider configurations. If the
+// the chain is set to dYdX, the dYdX node URL is required. We do so to ensure that the oracle is
+// always started using the oracle config that is expected to be stored by the chain.
+func createOracleConfig() error {
 	// If the providers is not empty, filter the providers to include only the
 	// providers that are specified.
-	providersFlag := *providersOpt
-	if providersFlag != "" {
+	if *chain == dYdXChain {
```
aljo242: could probably factor this out to a dydx specific function

cmd/config/main.go
```diff
@@ -236,123 +260,115 @@ var (
...
 // each provider as well as the custom conversion markets. We do so to ensure that the
 // oracle is always started using the market map that is expected to be stored by the
 // market map module.
-func createMarketMap() (mmtypes.MarketMap, error) {
+func createMarketMap() error {
+	if *chain == dYdXChain {
+		fmt.Fprintf(
+			os.Stderr,
+			"dYdX chain requires the use of a predetermined market map. please use the market map provided by the Skip/dYdX team or the default market map provided in /config/dydx/market.json",
+		)
```
aljo242: dydx specific function.

if we have this all factored out already, it will be easier to reason about in the future

cmd/config/main.go
```diff
@@ -383,23 +396,40 @@ func createMarketMap() (mmtypes.MarketMap, error) {
...
 
 	// Validate the market map.
 	if err := marketMap.ValidateBasic(); err != nil {
-		return mmtypes.MarketMap{}, fmt.Errorf("error validating the market map: %w", err)
+		return fmt.Errorf("error validating the market map: %w", err)
+	}
+
+	// Open the local market config file. This will overwrite any changes made to the
+	// local market config file.
+	f, err := os.Create(*marketCfgPath)
```
aljo242, line 407: Can we check if all flags were populated at the start of `main`?

Just so we can feel good about not having any `nil` dereferences.  It should be fine, but would be good to have the check in place

providers/apis/dydx/utils.go
```diff
@@ -51,10 +51,10 @@ var DefaultAPIConfig = config.APIConfig{
 	URL:              "localhost:1317",
 }
 
-// These are references to the different providers that are supported by the dYdX market map.
+// ProviderMapping are references to the different providers that are supported by the dYdX market map.
```
aljo242:
```suggestion
// ProviderMapping is references to the different providers that are supported by the dYdX market map.
```

aljo242 (approved):
lgtm with nits

cmd/config/main.go
davidterpay: I think `flags.Parse()` does this. If nothing exists it populates with the default value.

PR #261, opened 2024-03-21
fix: MM URL

davidterpay:


config/dydx/oracle.json
```diff
@@ -431,7 +431,7 @@
           "reconnectTimeout": 2000000000,
           "maxQueries": 1,
           "atomic": true,
-          "url": "https://dydx-api.polkachu.com",
+          "url": "localhost:1317",
```
Eric-Warehime, line 434: Is this some arbitrary localhost port?

davidterpay: its the rest endpoint for cosmos chains.

Eric-Warehime: Should we not provide a legitimate public API to fallback to as a default?

Eric-Warehime: > its the rest endpoint for cosmos chains.

Gotcha. I suppose it's fine to assume you're running the node locally then.

Eric-Warehime (approved):

PR #262, opened 2024-03-21
refactor: only validate VE size in `ValidateOracleVoteExtensions`

aljo242:
Closes BLO-958
Closes BLO-1003 

no longer check:
- id is in state
- price validity in state

Only check if NumPrices <= MaxNumCPs.  MaxNumCPs is derived from the number of currency pairs that were in state in the prev block.  This implicitly accounts for CPs being removed in that block.

Updates tests accordingly.


abci/ve/utils.go
```diff
@@ -26,23 +26,21 @@ func ValidateOracleVoteExtension(
 	ve vetypes.OracleVoteExtension,
 	strategy currencypair.CurrencyPairStrategy,
 ) error {
+	maxNumCP, err := strategy.GetMaxNumCP(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to get max price bytes size: %w", err)
+	}
```
g3ndo9kar9, line 32: nit: tests?

g3ndo9kar9 (approved):
lgtm

Eric-Warehime (approved):

PR #266, opened 2024-03-22
feat(`x/marketmap`): init types and validation

aljo242:
Closes BLO-1029


proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// MarketMap maps ticker strings to their Markets.
+message MarketMap {
+  option (gogoproto.goproto_stringer) = false;
+  option (gogoproto.stringer) = false;
+
+  // Markets is the full list of tickers and their associated configurations
+  // to be stored on-chain.
+  map<string, Market> markets = 1 [ (gogoproto.nullable) = false ];
```
davidterpay, line 83: Do we still need to have this be a map? i'd prefer moving this into a list for simplicity. All off-chain access to this will have to have its own transformations anyways so map doesn't introduce much ease.

aljo242: Yeah we can definitely do this

x/mm2/types/market.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+)
+
+// ValidateBasic performs aggregate validation for all fields in the MarketMap. We consider
+// the market map to be valid iff:
+//
+// 1. Each ticker a provider supports is included in the main set of tickers.
+// 2. Each ticker is valid.
+// 3. Each provider is valid.
+// 4. Aggregation function is valid.
```
davidterpay:
```suggestion
// ValidateBasic performs aggregate validation for all fields in the MarketMap. We consider
// the market map to be valid iff:
//
// 1. Each ticker is valid.
// 2. Each provider is valid.
// 3. Any provider configuration that requires a transformation (normalizeByPair) must be included in the main set of tickers.
```

x/mm2/types/market.go
```diff
@@ -0,0 +1,126 @@
...
+func (mm *MarketMap) ValidateIndexPriceAggregation() error {
+	for _, market := range mm.Markets {
+		if err := market.ValidateBasic(); err != nil {
+			return err
+		}
+
+		for _, providerConfig := range market.ProviderConfigs {
+			if providerConfig.NormalizeByPair != nil {
+				if _, found := mm.Markets[providerConfig.NormalizeByPair.String()]; !found {
+					return fmt.Errorf("provider index of %s was not found in the marketmap", providerConfig.NormalizeByPair.String())
```
davidterpay:
```suggestion
					return fmt.Errorf("provider's (%s) pair for normalization (%s) was not found in the marketmap", providerConfig.Name, providerConfig.NormalizeByPair.String())
```

davidterpay: or something like this, it can be hard pinpointing what provider was misconfigured otherwise.

proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+  // the ticker valid.
+  uint64 min_provider_count = 3;
+
+  // Enabled is the flag that denotes if the Ticker is enabled for price
+  // fetching by an oracle.
+  bool enabled = 14;
+
+  // MetadataJSON is a string of JSON that encodes any extra configuration
+  // for the given ticker.
+  string metadata_JSON = 15;
```
g3ndo9kar9, line 48: This is where we'd be storing the `MinPriceChangePPM`?

proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+  // OffChainTicker is the off-chain representation of the ticker i.e. BTC/USD.
+  // The off-chain ticker is unique to a given provider and is used to fetch the
+  // price of the ticker from the provider.
+  string off_chain_ticker = 2;
+
+  // NormalizeByPair is the currency pair for this ticker to be normalized by.
+  // For example, if the desired Ticker is BTD/USD, this market could be reached
+  // using: OffChainTicker = BTC/USDT NormalizeByPair = USDT/USD This field is
+  // optional and nullable.
+  slinky.types.v1.CurrencyPair normalizeByPair = 3;
```
g3ndo9kar9: nit: `normalize_by_pair`

proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+
+  // Invert is a boolean indicating if the BASE and QUOTE of the market should
+  // be inverted. i.e. BASE -> QUOTE, QUOTE -> BASE
+  bool invert = 4;
+
+  // MetadataJSON is a string of JSON that encodes any extra configuration
+  // for the given provider config.
+  string metadata_JSON = 15;
```
g3ndo9kar9, line 73: nit: 5

proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+  // decimal places is used to convert the price to a human-readable format.
+  uint64 decimals = 2;
+
+  // MinProviderCount is the minimum number of providers required to consider
+  // the ticker valid.
+  uint64 min_provider_count = 3;
+
+  // Enabled is the flag that denotes if the Ticker is enabled for price
+  // fetching by an oracle.
+  bool enabled = 14;
```
g3ndo9kar9, line 44: Where are these proto field IDs coming from?

proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+  option (gogoproto.goproto_stringer) = false;
+  option (gogoproto.stringer) = false;
+
+  // CurrencyPair is the currency pair for this ticker.
+  slinky.types.v1.CurrencyPair currency_pair = 1
+      [ (gogoproto.nullable) = false ];
+
+  // Decimals is the number of decimal places for the ticker. The number of
+  // decimal places is used to convert the price to a human-readable format.
+  uint64 decimals = 2;
```
g3ndo9kar9, line 36: How do these compare to the dydx decimals? Those can be negative, no?

proto/slinky/mm2/v1/market.proto
g3ndo9kar9: We're never indexing directly into the market-config? If so, moving to a slice makes the index operation O(n)

x/mm2/types/market.go
```diff
@@ -0,0 +1,126 @@
...
+
+// ValidateIndexPriceAggregation validates the market map configuration and its expected configuration for
+// this aggregator. In particular, this will
+//
+//  1. Ensure that the market map is valid (ValidateBasic). This ensures that each of the provider's
+//     markets are supported by the market map.
+//  2. Ensure that each path has a corresponding ticker.
+//  3. Ensure that each path has a valid number of operations.
+//  4. Ensure that each operation has a valid ticker and that the provider supports the ticker.
+func (mm *MarketMap) ValidateIndexPriceAggregation() error {
```
g3ndo9kar9: nit: is this name still relevant?

x/mm2/types/market.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if len(m.ProviderConfigs) != len(other.ProviderConfigs) {
+		return false
+	}
+
+	for i, providerConfig := range m.ProviderConfigs {
+		if !providerConfig.Equal(other.ProviderConfigs[i]) {
+			return false
+		}
+	}
```
g3ndo9kar9, line 111: are there tests for this file?

proto/slinky/mm2/v1/market.proto
aljo242: yes

proto/slinky/mm2/v1/market.proto
aljo242: I like putting metadata or flags at the "end" of the proto def. 

Field IDs 1-15 take the same amount of data to encode, so there really isnt any different.  It is more of a stylistic thing.  If we wanted to extend fields in the ticker, we could just 4 in the future which is slightly cuter lol

proto/slinky/mm2/v1/market.proto
aljo242: During validation, you need to construct a map to check everything, then when the struct is consumed by users, they will use it in  a map like fashion.  It is easier to leave as a map.

x/mm2/types/market.go
aljo242: yes - we are validating the market map for the index price aggregation strategy it must be set up to support

proto/slinky/mm2/v1/market.proto
davidterpay: we just take the absolute value in our case.

x/mm2/types/market.go
aljo242: Adding

davidterpay (approved):
lgtm after some testing is pushed

x/mm2/types/market.go
aljo242: removed and just using `ValidateBasic()`

proto/slinky/mm2/v1/market.proto
Eric-Warehime: 🤔 Not sure if we have anything written already utilizing this JSON/how the ser/de struct will be integrated into the sidecar that I could look at? 

I want to minimize the number of places we have to keep stateful logic around which chain we're running for, and this seems like a field that would potentially be different not on a per-provider basis but on a per-chain basis.

`MinPriceChangePPM` is a good example of that which is at the moment dydx-specific.

x/mm2/types/ticker.go
```diff
@@ -0,0 +1,68 @@
...
+	if t.MinProviderCount < DefaultMinProviderCount {
+		return fmt.Errorf("min provider count must be at least %d; got %d for %s", DefaultMinProviderCount, t.MinProviderCount, t.CurrencyPair.String())
+	}
+
+	if err := t.CurrencyPair.ValidateBasic(); err != nil {
+		return err
+	}
+
+	if len(t.Metadata_JSON) > MaxMetadataJSONFieldLength {
+		return fmt.Errorf("metadata json field is longer than maximum length of %d", MaxMetadataJSONFieldLength)
```
Eric-Warehime, line 55: Not covered by tests.

x/mm2/types/market.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// ValidateBasic validates the market map configuration and its expected configuration.
+//
+//	In particular, this will
+//
+//	1. Ensure that the market map is valid (ValidateBasic). This ensures that each of the provider's
+//	   markets are supported by the market map.
+//	2. Ensure that each path has a corresponding ticker.
+//	3. Ensure that each path has a valid number of operations.
```
Eric-Warehime: The concept of a "path" no longer exists right?

x/mm2/types/market.go
```diff
@@ -0,0 +1,116 @@
...
+//	4. Ensure that each operation has a valid ticker and that the provider supports the ticker.
```
Eric-Warehime: Is the `Operation` type gone now too?

x/mm2/types/market.go
```diff
@@ -0,0 +1,116 @@
...
+func (mm *MarketMap) ValidateBasic() error {
+	for _, market := range mm.Markets {
+		if err := market.ValidateBasic(); err != nil {
+			return err
+		}
+
+		for _, providerConfig := range market.ProviderConfigs {
+			if providerConfig.NormalizeByPair != nil {
+				if _, found := mm.Markets[providerConfig.NormalizeByPair.String()]; !found {
+					return fmt.Errorf("provider's (%s) pair for normalization (%s) was not found in the marketmap", providerConfig.Name, providerConfig.NormalizeByPair.String())
```
Eric-Warehime, line 23: Needs test.

x/mm2/types/market.go
```diff
@@ -0,0 +1,116 @@
...
+
+	seenProviders := make(map[string]struct{})
+	for _, providerConfig := range m.ProviderConfigs {
+		if err := providerConfig.ValidateBasic(); err != nil {
+			return err
+		}
+
+		// check for duplicate providers
+		if _, seen := seenProviders[providerConfig.Name]; seen {
+			return fmt.Errorf("duplicate provider found: %s", providerConfig.Name)
```
Eric-Warehime, line 61: Needs test.

x/mm2/types/market.go
aljo242: tested by "provider config includes a ticker that is not supported"

x/mm2/types/ticker.go
aljo242: tested by "invalid json len"

Eric-Warehime (approved):

proto/slinky/mm2/v1/market.proto
Christopher-Li, line 48: is there a reason why you skipped 4 - 13?

Also do you have any ideas or examples of what would be stored in here?

proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+  // optional and nullable.
+  slinky.types.v1.CurrencyPair normalize_by_pair = 3;
+
+  // Invert is a boolean indicating if the BASE and QUOTE of the market should
+  // be inverted. i.e. BASE -> QUOTE, QUOTE -> BASE
+  bool invert = 4;
+
+  // MetadataJSON is a string of JSON that encodes any extra configuration
+  // for the given provider config.
+  string metadata_JSON = 15;
```
Christopher-Li, line 73: skipped 5 - 14?

proto/slinky/mm2/v1/market.proto
```diff
@@ -0,0 +1,84 @@
...
+  // being set.
+  string name = 1;
+
+  // OffChainTicker is the off-chain representation of the ticker i.e. BTC/USD.
+  // The off-chain ticker is unique to a given provider and is used to fetch the
+  // price of the ticker from the provider.
+  string off_chain_ticker = 2;
+
+  // NormalizeByPair is the currency pair for this ticker to be normalized by.
+  // For example, if the desired Ticker is BTD/USD, this market could be reached
```
Christopher-Li, line 62: nit: typo `BTC/USD`

x/mm2/types/provider.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+
+	if len(pc.Metadata_JSON) > MaxMetadataJSONFieldLength {
+		return fmt.Errorf("metadata json field is longer than maximum length of %d", MaxMetadataJSONFieldLength)
+	}
+
+	if err := json.IsValid([]byte(pc.Metadata_JSON)); err != nil {
+		return fmt.Errorf("invalid provider config metadata json: %w", err)
+	}
```
Christopher-Li, line 32: should we add any validation that ProviderConfig.metadata_JSON holds any particular data?

x/mm2/types/ticker.go
```diff
@@ -0,0 +1,72 @@
...
+	slinkytypes "github.com/skip-mev/slinky/pkg/types"
+)
+
+const (
+	// DefaultMaxDecimals is the maximum number of decimals allowed for a ticker.
+	DefaultMaxDecimals = 36
+	// DefaultMinProviderCount is the minimum number of providers required for a
+	// ticker to be considered valid.
+	DefaultMinProviderCount = 1
+	// MaxMetadataJSONFieldLength is the maximum length of the MetadataJSON field.
```
Christopher-Li, line 18: nit: what unit is this? In bytes or bits?

proto/slinky/mm2/v1/market.proto
aljo242: There is no impact to the size of the data structure when choosing field IDs 1 - 16 (all take 2bz to encode).

This is just a stylistic move of putting metadata (an optional field) at the "end" of the struct

x/mm2/types/provider.go
aljo242: We can at a later time when we have solid metadata types we care about.  Now that we are adding defi providers like raydium, they use this field, so we can add those once merged.  Good callout.  adding a linear issue for us to track this

x/mm2/types/ticker.go
aljo242: bytes !

aljo242: updating comment !

PR #267, opened 2024-03-22
feat(`x/marketmap`): params type

aljo242:
Closes BLO-1030


x/mm2/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+	}, nil
+}
+
+// ValidateBasic performs stateless validation of the Params.
+func (p *Params) ValidateBasic() error {
+	if p.MarketAuthorities == nil {
+		return fmt.Errorf("cannot create Params with nil authority")
+	}
+
+	seenAuthorities := make(map[string]struct{}, len(p.MarketAuthorities))
```
g3ndo9kar9, line 38: we perform this check so frequently (uniqueness of items in an array), perhaps it's useful for us to make a generic helper in `/lib` and start using that

proto/slinky/mm2/v1/params.proto
```diff
@@ -0,0 +1,15 @@
...
+
+// Params defines the parameters for the x/marketmap module.
+message Params {
+  // MarketAuthorities is the list of authority accounts that are able to
+  // control updating the marketmap.
+  repeated string market_authorities = 1;
+
+  // Version is the schema version for the MarketMap data structure and query
+  // response.
+  uint64 version = 2;
```
g3ndo9kar9: We're putting the version in params? I thought we'd be putting it in the market-map? How else would we be correlating between the mm data-structure + the version?

proto/slinky/mm2/v1/params.proto
```diff
@@ -0,0 +1,15 @@
...
+}
```
g3ndo9kar9: nit: newline at EOF

proto/slinky/mm2/v1/params.proto
davidterpay: Can we have this be a string which is essentially just the tag?

x/mm2/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+)
+
+const (
+	// DefaultVersion is the default value for the Version Param.
+	DefaultVersion = 0
```
davidterpay: following up on the other point, we should just hard code this to be the tag when the mm is ready.

davidterpay: presumably `v1.0.0`

aljo242: What do you mean by this?

proto/slinky/mm2/v1/params.proto
aljo242: @davidterpay How would that be set? My thinking is that this was the schema version of the market map struct. Perhaps this isn't even a useful thing to have. We can do versioning things in the sidecar based on the tag

davidterpay: The marketmap scheme is likely not going to change very often but we should have a way of signaling validators when to update their side-car with a new tag/version. imo you kinda accomplish both in the same case, any schema change to the market map is going require a new side car version anyways

davidterpay: im assuming once this is ready we just set the default to v1.0.0

x/mm2/types/params.go
davidterpay: the release we cut with the mm is probably going to be v1.0.0

aljo242:
@nivasan1 @davidterpay removing `Version` field for now until we get alignment on how / if we will use this

davidterpay (approved):

x/mm2/types/params.go
aljo242: https://linear.app/skip/issue/BLO-1059/create-lib-function-for-checking-uniqueness-of-items-in-a-list

proto/slinky/mm2/v1/params.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package slinky.mm2.v1;
+
+option go_package = "github.com/skip-mev/slinky/x/mm2/types";
+
+// Params defines the parameters for the x/marketmap module.
+message Params {
+  // MarketAuthorities is the list of authority accounts that are able to
+  // control updating the marketmap.
+  repeated string market_authorities = 1;
```
Eric-Warehime, line 10: Maybe we can do this in a followup:

I thought our idea was to have a 2 tiered permissions system in which participants could update the mm, and a single "admin" authority could add/remove participants.

x/mm2/types/params.go
```diff
@@ -0,0 +1,52 @@
...
+func DefaultParams() Params {
+	return Params{
+		MarketAuthorities: []string{DefaultMarketAuthority},
+	}
+}
+
+// NewParams returns a new Params instance.
+func NewParams(authorities []string) (Params, error) {
+	if authorities != nil {
+		return Params{}, fmt.Errorf("cannot create Params with nil authority")
```
Eric-Warehime: Not sure how this works, shouldn't this be `if authorities == nil` ?

proto/slinky/mm2/v1/params.proto
aljo242: yeah - I added this in a follow up https://github.com/skip-mev/slinky/pull/293

Eric-Warehime (approved):

PR #270, opened 2024-03-22
feat(`x/marketmap`): keeper methods

aljo242:
Closes BLO-1032


x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/skip-mev/slinky/x/mm2/types"
+)
+
+// Keeper is the module's keeper implementation.
+type Keeper struct {
+	cdc codec.BinaryCodec
+
+	// module authority
+	authority sdk.AccAddress
```
davidterpay, line 19: How is this different from the authorities defined in params?

x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,181 @@
...
+
+				if !has {
+					return fmt.Errorf("currency pair %s in provider config does not exist", providerConfig.NormalizeByPair.String())
+				}
+			}
+		}
+	}
+
+	return nil
+}
```
davidterpay: I think this function might be a bit of a footgun since any external dependency to Create or Update will have to call validate state afterwards. Do you think it makes more sense to just move this into the top level Create and Update instead and have this be a internal helper function?

x/mm2/keeper/keeper.go
aljo242: This is the authority in charge of params (typically gov).  Because of this, the gov params have the final say on who the market authorities are

x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,193 @@
...
+
+// validateState is called after keeper modifications have been made to the market map to verify that
+// the aggregate of all updates has led to a valid state.
+func (k *Keeper) validateState(ctx sdk.Context, updates []types.Market) error {
+	for _, market := range updates {
+		// check that all paths already exist in the keeper store:
+		for _, providerConfig := range market.ProviderConfigs {
+			if providerConfig.NormalizeByPair != nil {
+				has, err := k.markets.Has(ctx, types.TickerString(providerConfig.NormalizeByPair.String()))
```
davidterpay: Given that you are checking whether a given market exists, its probably useful to implement this as a helper function that can also be exported in case we need it.

x/mm2/keeper/keeper_test.go
```diff
@@ -0,0 +1,182 @@
...
+	})
+
+	s.Run("get all tickers", func() {
+		got, err := s.keeper.GetAllMarketsMap(s.ctx)
+		s.Require().NoError(err)
+
+		s.Require().Equal(len(markets), len(got))
+		s.Require().Equal(marketsMap, got)
+	})
+}
```
davidterpay: Can you add test coverage for:
1. Market that has a normalize by pair that does not exist in state.
2. Updating a single market.
3. Updating a market that then includes a normalize by pair that does not exist.

davidterpay (commented):
I think these changes are good once we add a bit more code cov

x/mm2/keeper/keeper_test.go
aljo242: https://github.com/skip-mev/slinky/pull/270/commits/f10c268149233c5ed6b76981f5a9f91d3148804a

x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,203 @@
...
+		markets:     collections.NewMap(sb, types.MarketsPrefix, "markets", types.TickersCodec, codec.CollValue[types.Market](cdc)),
+		lastUpdated: collections.NewItem[uint64](sb, types.LastUpdatedPrefix, "last_updated", types.LastUpdatedCodec),
+		params:      params,
+	}
+}
+
+// SetLastUpdated sets the lastUpdated field to the current block height.
+func (k *Keeper) SetLastUpdated(ctx sdk.Context, height uint64) error {
+	return k.lastUpdated.Set(ctx, height)
+}
```
davidterpay, line 56: Is height required here if we are setting it to the current height?

davidterpay (approved):

x/mm2/keeper/keeper.go
Eric-Warehime: I still think we want 2 authorities in params and 1 here.

Keeper authority: controls who the admin is
Admin auth: controls who the participants are
Participants: Can update the market map

x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,203 @@
...
+	// Check if Ticker already exists for the provider
+	alreadyExists, err := k.markets.Has(ctx, types.TickerString(market.Ticker.String()))
+	if err != nil {
+		return err
+	}
+	if !alreadyExists {
+		return types.NewMarketDoesNotExistsError(types.TickerString(market.Ticker.String()))
+	}
+	// Create the config
+	return k.markets.Set(ctx, types.TickerString(market.Ticker.String()), market)
```
Eric-Warehime, line 115: No validation needs to happen on which fields are being updated?

x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,203 @@
...
+func (k *Keeper) CreateMarkets(ctx sdk.Context, markets []types.Market) error {
+	for _, market := range markets {
+		if err := k.createMarket(ctx, market); err != nil {
+			return err
+		}
+	}
+
+	if err := k.validateState(ctx, markets); err != nil {
+		return err
+	}
```
Eric-Warehime: I understand the need to do validation after all updates have been made, but won't the fact that markets have been created already leave us in a weird state? We'll have created a bunch of markets and errored out but the last updated height won't be changed and we'll have no way to roll back to the previous valid state.

aljo242: No - this is a transaction. So the update to state is all or nothing. There won't be any weird dangling intermediate states

x/mm2/keeper/keeper.go
aljo242: What kind of validation do you mean?

Eric-Warehime: I guess we don't need any? My initial thought was something along the lines of "which fields don't we want to make mutable". But I guess just the ticker's currency pair.

Eric-Warehime (approved):

x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return nil
+}
+
+// IsMarketValid checks if a market is valid by statefully checking if each of the currency pairs
+// specified by its provider configs are valid and in state.
+func (k *Keeper) IsMarketValid(ctx sdk.Context, market types.Market) error {
+	// check that all markets already exist in the keeper store:
+	for _, providerConfig := range market.ProviderConfigs {
```
Christopher-Li, line 144: can we reuse some of the `ValidateBasic` logic here?

x/mm2/keeper/keeper.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// GetMarket returns a market from the store by its currency pair string ID.
+func (k *Keeper) GetMarket(ctx sdk.Context, tickerStr string) (types.Market, error) {
+	return k.markets.Get(ctx, types.TickerString(tickerStr))
+}
+
+// GetAllMarkets returns the set of Market objects currently stored in state
+// as a map[TickerString] -> Markets.
+func (k *Keeper) GetAllMarkets(ctx sdk.Context) (map[string]types.Market, error) {
```
Christopher-Li, line 70: how does this return type differ from `collections.Map[types.TickerString, types.Market]`

aljo242: So, a `collections.Map` stores data in the KVstore in a map-like fashion. The access pattern is a `Get` which returns a single item, or to iterate through the collection which would let you return a list, or map.  Here, we return a map.

x/mm2/keeper/keeper.go
aljo242: Not here.  This is some additional stateful validation that requires us to use the keeper.  We need to verify that for each new market added, if they are Normalizing by another market (normalizeBy), does that normalizeBy exist in state.  Any other stateless validation will already have been done by now by `ValidateBasic()`.

aljo242: This protects us from the case where in state we have:
- BTC/USD
- MOG/USD

and we want to add a new market which is:
- ETH/USD (fed by an ETH/USDT feed, normalized by USDT/USD)

Only at a stateful verification stage, can we verify that the required USDT/USD exists.

PR #272, opened 2024-03-25
feat: Rename oracle to slinky, refactor mm provider factory

Eric-Warehime:
This PR renames `./build/oracle => ./build/slinky`.

It also refactors the marketmap provider factory to be more in line with the other factories (websocket and API). The idea being that once we remove the `aggregator` bit, we can get rid of all of the special casing for dYdX and rely on the factory to construct the proper marketmap provider for any chain we support.

I've pointed this PR at the `eric/convert-everything-to-cobra` branch since it made sense to change these in the context of the cobra refactor.


providers/factories/oracle/marketmap.go
```diff
@@ -33,49 +34,21 @@ func DefaultMarketMapProvider(
...
-		apiMetrics,
-	)
-	if err != nil {
-		return nil, err
+	switch cfg.Name {
+	case dydx.Name:
+		apiDataHandler, err = dydx.NewAPIHandler(cfg.API)
+		ids = []types.Chain{{ChainID: dydx.ChainID}}
+	default:
+		return nil, fmt.Errorf("unknown provider: %s", cfg.Name)
```
aljo242: This case should return the default marketmap API handler that hits `x/marketemap` right?

or is the idea that we'll fill this info out later after the refactor?

cmd/slinky/main.go
```diff
@@ -132,6 +132,10 @@ func runOracle() error {
 		orchestrator.WithMarketMap(marketCfg),
 		orchestrator.WithPriceAPIQueryHandlerFactory(oraclefactory.APIQueryHandlerFactory),             // Replace with custom API query handler factory.
 		orchestrator.WithPriceWebSocketQueryHandlerFactory(oraclefactory.WebSocketQueryHandlerFactory), // Replace with custom websocket query handler factory.
+		orchestrator.WithMarketMapperFactory(oraclefactory.MarketMapProviderFactory),
```
davidterpay, line 148: We need the marketmap provider factory to only be included in the dYdX options because it will either run the mm provider and update based on what is on dYdX main-net or it will fail to be constructed if the market map provider is not included in the oracle config. There is probably a cleaner way with dealing how to run this based per chain and whether the mm provider is desired. But I'm assuming this isnt desired behavior.

davidterpay (commented):
given that the mm provider is only going to be utilized by dYdX for the time being, we need to move back into dYdXOptions

providers/factories/oracle/marketmap.go
```diff
@@ -33,49 +34,21 @@ func DefaultMarketMapProvider(
...
-		requestHandler,
-		apiDataHandler,
-		apiMetrics,
-	)
-	if err != nil {
-		return nil, err
+	switch cfg.Name {
+	case dydx.Name:
+		apiDataHandler, err = dydx.NewAPIHandler(cfg.API)
+		ids = []types.Chain{{ChainID: dydx.ChainID}}
```
Eric-Warehime, line 44: I'm not sure how this is used right now, but it feels like we don't really need the chain-id as long as we point the mm provider at the correct node?

Eric-Warehime:
Approach I ended up with:
* Ignores aggregation type entirely. Run the full aggregation paths validation on all configs we get.
* Update CLIs as follows:
Config is pretty much as before.
```
slinky-config --chain "" -- node-http-url ""
```


Slinky requires a valid oracle.json path. Has optional flags, `market-config-path` or `market-config-update-path` which are mutually exclusive.   
`market-config-path` is the path to the `market.json` to start the system with (same as we currently have)--though that will get updated if you have a market map provider present.  
`market-config-update-path` is a place where we will write market updates in line with what the sidecar is doing. Sidecar initially bootstraps with an empty config and writes updates to this path.  
Specifying neither of these will initialize with an empty market map and updates will never be written to disk.
```
slinky --oracle-config-path "./oracle.json"
slinky --oracle-config-path "./oracle.json" --market-config-path ./market.json
slinky --oracle-config-path "./oracle.json" --market-config-update-path ./new-file.json
```

cmd/slinky/main.go
g3ndo9kar9: I'm confused why this even needs to be an option? Shouldn't we be able to have casing logic like websocket / api, and have the default be a nop updater (no mm-provider? Or a mm provider that never updates)

cmd/slinky/main.go
```diff
@@ -129,29 +123,33 @@ func runOracle() error {
 	}
 
 	metrics := oraclemetrics.NewMetricsFromConfig(cfg.Metrics)
+	aggregator, err := oraclemath.NewMedianAggregator(
```
g3ndo9kar9, line 133: I thought we agreed to have a single aggregation strat in the oracle (the dydx one)

providers/factories/oracle/marketmap.go
g3ndo9kar9: ya, how would our configs work if we have a market-map for multiple chains? Have we even resolved this?

aljo242 (approved):

cmd/slinky-config/main.go
```diff
@@ -422,13 +424,27 @@ func createMarketMap() error {
 			providers := tickersToProviders[tickerStr].Providers
 			providers = append(providers, config)
 			tickersToProviders[tickerStr] = mmtypes.Providers{Providers: providers}
+
+			if _, ok := tickersToPaths[tickerStr]; !ok {
+				tickersToPaths[tickerStr] = mmtypes.Paths{}
+			}
+			paths := tickersToPaths[tickerStr].Paths
+			paths = append(paths, mmtypes.Path{Operations: []mmtypes.Operation{
```
g3ndo9kar9, line 432: We're medianizing over all paths, no? This would give extra weight to the raw feed if it alr existed?

cmd/slinky-config/main.go
```diff
@@ -222,77 +223,77 @@ func init() {
...
 		"",
 		"",
-		"http endpoint of the cosmos sdk node corresponding to the chain id (typically something like localhost:1317). this should only be specified if required by the chain.",
+		"Http endpoint of the cosmos sdk node corresponding to the chain (typically localhost:1317 or a remote API). This should only be specified if required by the chain.",
 	)
 	rootCmd.Flags().StringVarP(
 		&host,
 		"host",
 		"",
 		"0.0.0.0",
```
g3ndo9kar9, line 253: Are we checking that the default value isn't used if no chain is given here?

pkg/math/oracle/median.go
```diff
@@ -38,11 +38,6 @@ func NewMedianAggregator(
 		return nil, err
 	}
 
-	if cfg.AggregationType != mmtypes.AggregationType_INDEX_PRICE_AGGREGATION {
```
g3ndo9kar9, line 41: Is there a reason we can't remove the agg-type from the config, now?

cmd/slinky/main.go
Eric-Warehime: This is the dydx aggregation.

cmd/slinky-config/main.go
Eric-Warehime: It's just a pure median over all feeds with optional 1 hop inversion.

pkg/math/oracle/median.go
Eric-Warehime: I left it since we're rewriting all of the mm stuff and deleting all the proto and tests didn't seem worth it at this point.

cmd/slinky-config/main.go
Eric-Warehime: This variable is only used if dydx chain is specified otherwise it's ignored at the moment.

oracle/constants/chains.go
```diff
@@ -1,25 +1,3 @@
...
-		Name: "dYdX Mainnet",
-		ID:   "dydx-mainnet-1",
-	}
-	// DYDXTestnet is the chain id for the testnet.
-	DYDXTestnet = Chain{
-		Name: "dYdX Testnet",
-		ID:   "dydx-testnet-4",
-	}
-)
+const Dydx = "dydx"
```
davidterpay:
```suggestion
const DYDX = "dydx"
```

small nit

pkg/math/oracle/median.go
davidterpay: What is the reason we need to remove this rn? From my perspective I put this in the case where operators accidentally use the wrong market.json file. this was an easy way to make sure things were configured correctly. the dYdX market params API constructs the market.json to include the field to be set correctly.

providers/factories/oracle/marketmap.go
davidterpay: Yea this is currently hard coded based on the desired chain but it doesnt really matter since all you have to do is point it to the correct URL. The providers were purposfully designed to assume you have inputs when constructing URLs, but in the future we can probably make it allow static or no entries.

davidterpay (commented):
We're going to need to update the local market.json if we want to enforce that the single aggregation type is utilized otherwise any build is going to fail on start.

Eric-Warehime:
> We're going to need to update the local market.json

@davidterpay It's updated in this PR

cmd/slinky/main.go
davidterpay: This is an option that lets the orchestrator know how to construct a market map provider if one is included in the oracle.json. Reading through this again I think this looks good.

davidterpay: The factories take care of casing based on the API/Websocket implementation

pkg/math/oracle/median.go
Eric-Warehime: Supporting multiple aggregation types seems like a really bad idea in general. It will make it even harder to reason about how a price was determined and introduce potential for multi-modal behavior in our on chain prices based on a bug in one place but not another.

I also removed it because it was the one stateful piece of validation that was preventing us from bootstrapping the oracle without a `market.json` config.

cmd/slinky-config/main.go
```diff
@@ -222,77 +223,77 @@ func init() {
...
-		"flag that enables debug mode. specifically the side-car will run in debug mode. this is useful for local development / debugging.",
+		"Flag that enables debug mode for the side-car. This is useful for local development / debugging.",
 	)
 	rootCmd.Flags().DurationVarP(
 		&updateInterval,
 		"update-interval",
 		"",
 		500*time.Millisecond,
-		"interval at which the oracle will update the prices. this should be set to the interval desired by the chain.",
+		"Interval at which the oracle will update the prices. This should be set to the interval desired by the chain.",
```
davidterpay, line 289: I would make sure that this is set to 500 milliseconds

davidterpay (approved):

pkg/math/oracle/median.go
davidterpay: yea agree, multiple not gud. im fine making this change now

PR #273, opened 2024-03-25
feat(`x/marketmap`): tx and msg service

aljo242:
Closes BLO-1033


x/mm2/keeper/msg_server.go
```diff
@@ -0,0 +1,136 @@
...
+		// TODO hooks
+
+		// TODO events
+	}
+
+	// validate that the new state of the marketmap is valid
+	err = ms.k.validateState(ctx, msg.CreateMarkets)
+	if err != nil {
+		return nil, fmt.Errorf("invalid state resulting from update: %w", err)
+	}
```
davidterpay: Why is this not just calling the CreateMarkets function? Is this just not up to date?

x/mm2/keeper/msg_server.go
```diff
@@ -0,0 +1,136 @@
...
+		// TODO events
+	}
+
+	// validate that the new state of the marketmap is valid
+	err = ms.k.validateState(ctx, msg.UpdateMarkets)
+	if err != nil {
+		return nil, fmt.Errorf("invalid state resulting from update: %w", err)
+	}
+
+	return &types.MsgUpdateMarketsResponse{}, ms.k.SetLastUpdated(ctx, uint64(ctx.BlockHeight()))
```
davidterpay: Same applies here.

x/mm2/keeper/msg_server.go
```diff
@@ -0,0 +1,136 @@
...
+
+	params, err := ms.k.GetParams(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get marketmap params: %w", err)
+	}
+
+	found := checkMarketAuthority(msg.Authority, params)
+	if !found {
+		return nil, fmt.Errorf("request signer %s does not match module market authorities", msg.Authority)
+	}
```
davidterpay, line 41: Just a sanity check (im sure ive asked this before), but are we certain that signers can't just set this to an authority address when they create the transaction but sign with a different key?

x/mm2/keeper/msg_server_test.go
```diff
@@ -0,0 +1,218 @@
...
+			btcusdt.Ticker.String(): btcusdt,
+			usdtusd.Ticker.String(): usdtusd,
+		},
+	})
+
+	// TODO: test hooks
+
+	s.Run("unable to process for invalid authority", func() {
+		msg = &types.MsgCreateMarkets{
+			Authority: "invalid",
```
davidterpay, line 40: Can we add a separate test case that inputs a valid bech32 address?

x/mm2/keeper/msg_server_test.go
```diff
@@ -0,0 +1,218 @@
...
+	// set a market in the map
+	s.Run("unable to process nil request", func() {
+		resp, err := msgServer.UpdateMarkets(s.ctx, nil)
+		s.Require().Error(err)
+		s.Require().Nil(resp)
+	})
+
+	s.Run("able to update market that already exists", func() {
+		tickerUpdate := btcusdt
+		tickerUpdate.Ticker.Decimals = 1
```
davidterpay, line 154: I would bring this up to the team, but do we anticipate that there are certain things that should not be able to change after the market is initially created?

davidterpay (commented):
Left some comments

x/mm2/keeper/msg_server.go
aljo242: no, we shouldn't call that here because in the msg server we still want to have extra things happen like events and hooks to be triggered (but only in the case of the msg server)

perhaps create-markets should also not be exported.

x/mm2/keeper/msg_server.go
aljo242: no, we have the proto-annotation that this must be the signer of the tx.

Same way any gov signed message is verified

x/mm2/keeper/msg_server_test.go
aljo242: The only one that I can really think of is that the currency pair does not change

aljo242: which is already enforced

davidterpay (approved):

proto/slinky/mm2/v1/tx.proto
```diff
@@ -0,0 +1,80 @@
...
+import "gogoproto/gogo.proto";
+import "cosmos_proto/cosmos.proto";
+import "amino/amino.proto";
+import "slinky/mm2/v1/market.proto";
+import "slinky/mm2/v1/params.proto";
+
+option go_package = "github.com/skip-mev/slinky/x/mm2/types";
+
+// Msg is the message service for the x/marketmap module.
+service Msg {
```
Eric-Warehime, line 14: I think there's some stuff to add around permissions which I've mentioned in other PRs. Like add/remove market participant/mapper or whatever we call them.

x/mm2/keeper/msg_server.go
Eric-Warehime: I'm confused about this as well. This seems to be the same logic that the CreateMarkets call is doing, same w/ UpdateMarkets I believe.

aljo242:
@Eric-Warehime @davidterpay i removed the `CreateMarkets` and `UpdateMarkets` functions.  It didnt make sense to have them, as there are cases where we want to just solely set a market (genesis, etc) and cases where we want to set, run hooks, emit events, -> validate (msg server).

So the only keeper methods i am exposing give you the building blocks for all of that, but nothing more.

davidterpay:
why cant the create / update market keeper method just encapsulate calling the hook, setting the latest height it was updated, and validating state? imo those operations should be atomic from the perspective of what create market does?

proto/slinky/mm2/v1/tx.proto
```diff
@@ -0,0 +1,80 @@
...
+  option (cosmos.msg.v1.service) = true;
+
+  // CreateMarkets creates markets from the given message.
+  rpc CreateMarkets(MsgCreateMarkets) returns (MsgCreateMarketsResponse);
+
+  // UpdateMarkets updates markets from the given message.
+  rpc UpdateMarkets(MsgUpdateMarkets) returns (MsgUpdateMarketsResponse);
+
+  // Params defines a method for updating the x/marketmap module parameters.
+  rpc Params(MsgParams) returns (MsgParamsResponse);
```
Christopher-Li, line 24: shouldnt this be named UpdateParams?

x/mm2/keeper/msg_server.go
```diff
@@ -0,0 +1,136 @@
...
+		// TODO hooks
+
+		// TODO events
+	}
+
+	// validate that the new state of the marketmap is valid
+	err = ms.k.ValidateState(ctx, msg.CreateMarkets)
+	if err != nil {
+		return nil, fmt.Errorf("invalid state resulting from update: %w", err)
+	}
```
Christopher-Li, line 59: do we need to delete the newly created markets?

x/mm2/keeper/msg_server.go
```diff
@@ -0,0 +1,136 @@
...
+		// TODO hooks
+
+		// TODO events
+	}
+
+	// validate that the new state of the marketmap is valid
+	err = ms.k.ValidateState(ctx, msg.UpdateMarkets)
+	if err != nil {
+		return nil, fmt.Errorf("invalid state resulting from update: %w", err)
+	}
```
Christopher-Li, line 98: same comment as above

x/mm2/keeper/msg_server.go
aljo242: No - any msg server level function in cosmos sdk is a "transaction", so its "all or nothing" like a db transaction.

The state is only applied if the entire tx is valid

PR #275, opened 2024-03-25
feat(`x/marketmap`): query server

aljo242:
Closes BLO-1034


davidterpay (approved):

Eric-Warehime (approved):

proto/slinky/mm2/v1/query.proto
```diff
@@ -0,0 +1,85 @@
...
+  rpc LastUpdated(LastUpdatedRequest) returns (LastUpdatedResponse) {
+    option (google.api.http).get = "/slinky/mm2/v1/last_updated";
+  }
+
+  // Params returns the current x/marketmap module parameters.
+  rpc Params(ParamsRequest) returns (ParamsResponse) {
+    option (google.api.http) = {
+      get : "/slinky/mm2/v1/params"
+    };
+  }
```
Christopher-Li, line 36: nit: what is mm2?

aljo242: This is the "temp" name we are naming this module. We already had a v0 version of the module named x/marketmap.

The plan is to replace that module with this one we are developing now. So the mm2 is just a temp name until we do the full switch over

proto/slinky/mm2/v1/query.proto
```diff
@@ -0,0 +1,85 @@
...
+  // and markets.
+  MarketMap market_map = 1 [ (gogoproto.nullable) = false ];
+
+  // LastUpdated is the last block height that the market map was updated.
+  // This field can be used as an optimization for clients checking if there
+  // is a new update to the map.
+  uint64 last_updated = 2;
+
+  // ChainId is the chain identifier for the market map.
+  string chain_id = 3;
```
Christopher-Li, line 55: could you elaborate on this? What does this do?

proto/slinky/mm2/v1/query.proto
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// MarketRequest is the query request for the Market query.
+// It takes the currency pair of the market as an argument.
+message MarketRequest {
+  // CurrencyPair is the currency pair associated with the market being
+  // requested.
+  slinky.types.v1.CurrencyPair currency_pair = 3
+      [ (gogoproto.nullable) = false ];
```
Christopher-Li, line 64: 3?

proto/slinky/mm2/v1/query.proto
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  // Market returns a market stored in the x/marketmap
+  // module.
+  rpc Market(MarketRequest) returns (MarketResponse) {
+    option (google.api.http).get = "/slinky/mm2/v1/market";
+  }
+
+  // LastUpdated returns the last height the market map was updated at.
+  rpc LastUpdated(LastUpdatedRequest) returns (LastUpdatedResponse) {
```
Christopher-Li, line 27: why do we need a separate query for last updated when marketmap query exists?

aljo242: This query has significantly lower state read overhead (O(1)). 

To get the market map from state, it is O(N). for N markets

Clients can be configured to only poll the `LastUpdated` field, and if it is different from some value they have cached, only then do they need the full query. 

Just a potential optimization.

proto/slinky/mm2/v1/query.proto
aljo242: this was a mistake - good catch

proto/slinky/mm2/v1/query.proto
Christopher-Li: If I understand correctly, before this code is shipped to imported into our code base, you will be replacing `mm2`?

aljo242: correct

PR #276, opened 2024-03-25
feat(`x/marktmap`): keeper genesis

aljo242:
`InitGenesis` and `ExportGenesis`

Closes BLO-1042


x/mm2/keeper/genesis.go
```diff
@@ -0,0 +1,60 @@
...
+	if err := k.SetParams(ctx, gs.Params); err != nil {
+		panic(err)
+	}
+
+	// TODO hooks
+}
+
+// ExportGenesis retrieves the genesis from state.
+func (k *Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
+	markets, err := k.GetAllMarketsMap(ctx)
```
Eric-Warehime: 😆 This function name reads pretty awkwardly.

Eric-Warehime: Wish it could be `GetAllMarkets`.

x/mm2/keeper/genesis.go
```diff
@@ -0,0 +1,60 @@
...
+	// validate the genesis
+	if err := gs.ValidateBasic(); err != nil {
+		panic(err)
+	}
+
+	for _, market := range gs.MarketMap.Markets {
+		if err := k.createMarket(ctx, market); err != nil {
+			panic(err)
+		}
+	}
```
davidterpay: Why not `CreateMarkets`? That does the extra state validation we probably want no?

x/mm2/keeper/genesis_test.go
```diff
@@ -0,0 +1,113 @@
...
+				usdcusd.Ticker.String(): usdcusd,
+			},
+		}
+
+		// assert that InitGenesis panics
+		s.Require().Panics(func() {
+			s.keeper.InitGenesis(s.ctx, *gs)
+			s.keeper.InitGenesis(s.ctx, *gs)
+		})
+	})
```
davidterpay, line 43: What is this testing?

davidterpay: Can we add a test that checks that the market map is correctly configured? i.e. normalizeByPair exists for all etc.

davidterpay (commented):
small nit, but otherwise lgtm

x/mm2/keeper/genesis.go
aljo242: the genesis will already be validated because we have all of the information needed.   We also want to set `LastUpdated` to the correct value set in the genesis

x/mm2/keeper/genesis_test.go
aljo242: we can't run init genesis twice.  this is because the markets are already set in state and you can't do that again

aljo242: added tests in previous PRs

davidterpay (approved):

Eric-Warehime (approved):

x/mm2/keeper/genesis.go
aljo242: it is now :)

PR #278, opened 2024-03-25
feat: Main Metrics Readme

davidterpay:


cmd/config/main.go
```diff
@@ -114,7 +114,7 @@ var (
 		// ----------------------Metrics Config-----------------------	//
 		// -----------------------------------------------------------	//
 		Metrics:        config.MetricsConfig{},
-		UpdateInterval: 1500 * time.Millisecond,
+		UpdateInterval: 500 * time.Millisecond,
```
davidterpay, line 117: lowering the interval since dYdX block time is less than 1s

contrib/images/slinky.sidecar.e2e.Dockerfile
```diff
@@ -16,7 +16,7 @@ EXPOSE 8080
 EXPOSE 8002
 
 COPY --from=builder /src/slinky/build/* /usr/local/bin/
-COPY --from=builder /src/slinky/config/local /etc/slinky/default_config
```
aljo242: i feel like this should be a `dydx` specifc image.  We should still maintain a default one

metrics.md
```diff
@@ -0,0 +1,226 @@
...
+# Metrics
+
+> **Definitions**:
+>
+> * **Market**: A market is a pair of assets that are traded against each other. For example, the BTC-USD market is the market where Bitcoin is traded against the US Dollar.
+> * **Price Feed**: A price feed is indexed by a price provider and a market. For example, the Coinbase API provides a price feed for the BTC-USD market.
+> * **Price Provider**: A price provider is a service that provides price data for a given market. For example, the Coinbase API is a price provider for the Coinbase markets.
+> * **Market Map Provider**: A market map provider is a service that supplies the markets that the side-car needs to fetch data for.
+
+The side-car exposes metrics on the `/metrics` endpoint. These metrics are in the Prometheus format and can be scraped by Prometheus or any other monitoring system that supports Prometheus format.
```
aljo242: is there a default port this is exposed on?

metrics.md
```diff
@@ -0,0 +1,226 @@
+# Side-Car Metrics
```
aljo242: Instead of everything being prefixed would "side-car"

Why don't we just use "Slinky"?

metrics.md
```diff
@@ -0,0 +1,226 @@
...
+* [`side_car_health_check_ticker_updates_total`](#side_car_health_check_ticker_updates_total): This metric is a counter that increments every time the side-car updates the price of a given market. This is a good indicator of the overall health of a given market.
+* [`side_car_health_check_provider_updates_total`](#side_car_health_check_provider_updates_total): This metric is a counter that increments everytime the side-car utilizes a given providers market data. This is a good indicator of the health of a given provider. Note that providers may not be responsible for every market. However, the side-car correctly tracks the number of expected updates for each provider.
+
+### `side_car_health_check_system_updates_total`
+
+This metric should be monotonically increasing. Specifically, the rate of this metric should be inversely correlated to the configured `UpdateInterval` in the oracle side-car configuration (`oracle.json`). To check this, you can run the following query in Prometheus:
```
aljo242: instead of "should be increasing"

I think we should use "is monotonically increasing"

There is no case where it is decreasing.

Same with below

metrics.md
```diff
@@ -0,0 +1,226 @@
...
+
+```promql
+rate(side_car_health_check_provider_updates_total{provider="coinbase_api", success="true"}[5m])
+```
+
+![Architecture Overview](./resources/side_car_health_check_provider_updates_total_rate.png)
+
+### Health Metrics Summary
+
+In summary, the health metrics should be monitored to ensure that the side-car is updating its internal state, updating the price of each market, and fetching data from the price providers as expected. The rate of updates for each of these metrics should be inversely correlated with the `UpdateInterval` in the oracle side-car configuration.
```
aljo242: Maybe we should describe how you could know there is something wrong here.

For example,

In the case where a health check update counter is not being updated at the rate it had been over X window of time...

metrics.md
```diff
@@ -0,0 +1,226 @@
...
+
+This metric represents the last recorded price for a given price feed. The metric is indexed by the provider and market. For example, if we want to check the last recorded price of the BTC-USD market from the Coinbase API, we can run the following query in Prometheus:
+
+```promql
+side_car_provider_price{provider="coinbase_api", id="btc/usd"}
+```
+
+![Architecture Overview](./resources/side_car_provider_price_coinbase.png)
+
+Alternatively, if we wanted to check that last recorded price of the BTC-USD market across all price providers, we can run the following query in Prometheus:
```
aljo242:
```suggestion
Alternatively, if we wanted to check that last recorded price of the BTC-USD market aggregated across all price providers, we can run the following query in Prometheus:
```

aljo242: or is this not true?

aljo242: how does this differ from `side_car_aggregated_price`?

metrics.md
davidterpay: this is the default that prom sets up whenever you run a prom server

davidterpay: but i think the port we expose should be explicitly defined, ill update

docker-compose.yml
```diff
@@ -10,14 +10,15 @@ services:
       context: .
       dockerfile: contrib/images/slinky.sidecar.e2e.Dockerfile
     volumes:
-      - ./config/local/oracle.json:/oracle/oracle.json
-      - ./config/local/market.json:/oracle/market.json
+      - ./config/dydx/oracle.json:/oracle/oracle.json
+      - ./config/dydx/market.json:/oracle/market.json
```
Eric-Warehime, line 14: What is the docker-compose used in again? e2e tests?

metrics.md
davidterpay: updated, this per provider price feeds.

metrics.md
```diff
@@ -0,0 +1,226 @@
...
+    * [Health Metrics](#health-metrics)
+    * [Prices Metrics](#prices-metrics)
+        * [Price Feed Metrics](#price-feed-metrics)
+        * [Aggregated Price Metrics](#aggregated-price-metrics)
+    * [HTTP Metrics](#http-metrics)
+    * [WebSocket Metrics](#websocket-metrics)
+
+# Metrics
+
+> **Definitions**:
```
Eric-Warehime, line 20: 🔥

metrics.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Health Metrics
+
+There are three primary health metrics that are exposed by Slinky:
+
+* [`side_car_health_check_system_updates_total`](#side_car_health_check_system_updates_total): This metric is a counter that increments every time the side-car updates its internal state. This is a good indicator of the side-car's overall health.
+* [`side_car_health_check_ticker_updates_total`](#side_car_health_check_ticker_updates_total): This metric is a counter that increments every time the side-car updates the price of a given market. This is a good indicator of the overall health of a given market.
+* [`side_car_health_check_provider_updates_total`](#side_car_health_check_provider_updates_total): This metric is a counter that increments everytime the side-car utilizes a given providers market data. This is a good indicator of the health of a given provider. Note that providers may not be responsible for every market. However, the side-car correctly tracks the number of expected updates for each provider.
+
+### `side_car_health_check_system_updates_total`
```
aljo242, line 41: I feel like these should all be `slinky_health_check...` 

instead of sidecar

aljo242 (approved):

metrics.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+![Architecture Overview](./resources/side_car_health_check_ticker_updates_total_rate.png)
+
+### `side_car_health_check_provider_updates_total`
+
+This metric should be monotonically increasing for each (provider, market) pair. To verify that the rate of updates for each provider is as expected, you can run the following query in Prometheus:
+
+```promql
+rate(side_car_health_check_provider_updates_total{provider="coinbase_api", success="true"}[5m])
```
Eric-Warehime, line 73: I'm a little confused about why the success rate over 5m is like 2/3 for all of the providers in these examples.

docker-compose.yml
davidterpay: nothing, just running locally

metrics.md
davidterpay: I'm going to update so that we use the rate described above and follow up with explanations on the examples.

g3ndo9kar9 (approved):
lgtm

PR #283, opened 2024-03-26
feat(`x/marketmap`): module definition

aljo242:
Closes BLO-1035


proto/slinky/mm2/module/v1/module.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package slinky.mm2.module.v1;
+
+import "cosmos/app/v1alpha1/module.proto";
+
+// Module is the config object of the builder module.
+message Module {
+  option (cosmos.app.v1alpha1.module) = {
+    go_import : "github.com/skip-mev/slinky/x/mm2"
```
Eric-Warehime, line 10: Is the plan to `s/mm2/marketmap/g` after we merge everything?

x/mm2/module.go
```diff
@@ -0,0 +1,257 @@
...
+		appmodule.Invoke(InvokeSetMarketMapHooks),
+	)
+}
+
+// Inputs contains the dependencies required for module construction.
+type Inputs struct {
+	depinject.In
+
+	// module dependencies
+	Config       *marketmapmodulev1.Module
```
Eric-Warehime, line 189: We're using the existing marketmap to construct the new marketmap? Is that correct?

proto/slinky/mm2/module/v1/module.proto
aljo242: yes.

x/mm2/module.go
aljo242: no - if you look at the import path - this is also from the new proto

Eric-Warehime: Ah I see. Just a confusing import name I guess. :O

Eric-Warehime (approved):

davidterpay (approved):

PR #289, opened 2024-03-26
feat(providers/api): Introduce HTTP Request modification (header / body) [BLO-1043]

g3ndo9kar9:
## In This PR
- I modify the `RequestHandler.Do` method to take an additional param, an `io.Reader` meant to represent the HTTP request body
    - This will be used for defi-providers that access node JSON-RPC APIs (we currently support only REST)
- I wrap the `APIDataHandler` abstraction with an `APIDataHandlerWithBody` abstraction, which introduces a new `CreateBody` method to generate an HTTP request body for a given request
    - This logic should be gracefully handled in `subTask`
### Let's discuss
- The wiring of errors -> metric error-codes is [messy](https://linear.app/skip/issue/BLO-1046/incorrect-handling-of-error-codes-metric-response-for-api-providers)


providers/base/api/handlers/api_query_handler_test.go
```diff
@@ -3,6 +3,7 @@ package handlers_test
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
```
aljo242: bai bai ioutil

providers/base/api/handlers/api_data_handler.go
```diff
@@ -23,3 +23,17 @@ type APIDataHandler[K providertypes.ResponseKey, V providertypes.ResponseValue]
...
+//
+//go:generate mockery --name APIDataHandlerWithBody --output ./mocks/ --case underscore
+type APIDataHandlerWithBody[K providertypes.ResponseKey, V providertypes.ResponseValue] interface {
+	APIDataHandler[K, V]
+
+	// CreateBody is used to create an HTTP response body that will be embdedded in the HTTP
+	// request to the server defined by the CreateURL response. The HTTP body generated will
+	// depend on the ids (identifiers of data to fetch) passed to the implementation.
+	CreateBody(ids []K) ([]byte, error)
+}
```
davidterpay, line 39: Why create an additional interface instead of just adding it to the other one? you already inherit `APIDataHandler` here. It should be relatively easy to update the existing API providers to just have no-ops for the `CreateBody` function.

davidterpay: I do like the additional method tho.

g3ndo9kar9: The outcomes are equivalent, this j ends up w/ less wasted code.

g3ndo9kar9: Doing things the other way, I'd have to change all the existing API providers w/ a method that is inherently unused, seems weird to me

providers/base/api/handlers/api_query_handler.go
```diff
@@ -210,11 +212,30 @@ func (h *APIQueryHandlerImpl[K, V]) subTask(
...
+
+			body = bytes.NewReader(bz)
+		}
+
 		// Make the request.
 		apiCtx, cancel := context.WithTimeout(ctx, h.config.Timeout)
 		defer cancel()
 
-		resp, err = h.requestHandler.Do(apiCtx, url)
+		resp, err = h.requestHandler.Do(apiCtx, url, body)
```
davidterpay, line 238:
```suggestion

		// get the request body if required for this API provider
		bz, err := apiHandler.CreateBody(ids)
        if err != nil {
            h.writeResponse(ctx, responseCh, providertypes.NewGetResponseWithErr[K, V](
                ids,
                providertypes.NewErrorWithCode(
                    errors.ErrCreateBodyWithErr(err),
                    providertypes.ErrorUnableToCreateRequestBody,
                )),
            )

            return nil
        }

		// Make the request.
		apiCtx, cancel := context.WithTimeout(ctx, h.config.Timeout)
		defer cancel()

		resp, err = h.requestHandler.Do(apiCtx, url, bytes.NewReader(bz))
```

davidterpay: i strongly prefer this over an additional interface

providers/base/api/metrics/api_utils.go
```diff
@@ -20,6 +20,8 @@ const (
 	UnexpectedStatusCode
 	// Success indicates that the provider successfully queried the data.
 	Success
+	// CreateBody.
```
davidterpay: nit: godoc

PR #299, opened 2024-03-29
feat(providers/api): Introduce a `APIPriceFetcher` interface [BLO-1026] [BLO-1047]

g3ndo9kar9:
## In This PR
- I introduce a `PriceFetcher` and have the `APIQueryHandler` depend on that interface
- This way, the logic executed in the `subTask` of each `APIQueryHandler` can be abstracted so that we can handle
    - REST API interactions
    - JSON-RPC client interactions (w/ custom client code): raydium / eth
    - GRPC client interactions: geyser for solana, etc
### Let's Discuss
- The metrics for the API providers is [messy](https://linear.app/skip/issue/BLO-1046/incorrect-handling-of-error-codes-metric-response-for-api-providers)
  - We should consolidate error-codes / error-types between the metrics pkg / providers pkg
- Our handling of the HTTP life-cycle is [incorrect](https://linear.app/skip/issue/BLO-1047/http-response-body-is-never-closed-in-api-query-handler)
   - Fixed in this PR
- Broadly speaking, our HTTP error code handling + metric reporting should be abstracted behind a shared `HTTPClient` interface (the `RequestHandler` currently does this, but does not accomplish much outside of wrapping the default `http.Client` impl
     - We shld remove the `RequestHandler` interface, and j introduce an `HTTPClient` interface
     - Move all of our HTTP error handling logic there
     - Introduce implementations that allow for fallback, retry, aggregation, etc + share them across all providers

### DIFF
--- a/providers/base/api/handlers/api_query_handler.go
+++ b/providers/base/api/handlers/api_query_handler.go
@@ -3,7 +3,6 @@ package handlers
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"strings"
 	"time"
 
@@ -13,7 +12,6 @@ import (
 
 	"github.com/skip-mev/slinky/oracle/config"
 	"github.com/skip-mev/slinky/pkg/math"
-	"github.com/skip-mev/slinky/providers/base/api/errors"
 	"github.com/skip-mev/slinky/providers/base/api/metrics"
 	providertypes "github.com/skip-mev/slinky/providers/types"
 )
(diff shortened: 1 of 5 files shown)


providers/base/api/handlers/api_query_handler.go
```diff
@@ -73,24 +78,20 @@ func NewAPIQueryHandler[K providertypes.ResponseKey, V providertypes.ResponseVal
...
-	}
-
 	if metrics == nil {
 		return nil, fmt.Errorf("no metrics specified for api query handler")
 	}
 
+	priceFetcher, err := NewRestAPIPriceFetcher(requestHandler, apiHandler, metrics, cfg, logger)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create price fetcher: %w", err)
+	}
```
davidterpay, line 88: just a note: this looks good to me but i anticipate that this may require some changes to the orchestrator implementation to account for the fact that the API query handler now utilizes a single interface with various implementations.

davidterpay: requestHandler RequestHandler,
this assumes rest atm but we can probably just accept a single price fetcher interface here instead of `RequestHandler` and `DataHandler`.

nivasan1: What is the requested change here?

g3ndo9kar9: good note

davidterpay (approved):

Eric-Warehime (approved):
Nice, LGTM

aljo242 (approved):

PR #300, opened 2024-04-01
fix: ABCI Removal Fix of Currency Pairs

davidterpay:
Let's walk through a few scenarios:


abci/strategies/currencypair/delta.go
```diff
@@ -120,11 +120,3 @@ func (s *DeltaCurrencyPairStrategy) getOnChainPrice(ctx sdk.Context, cp slinkyty
 	s.cache[cp] = currentPrice
 	return currentPrice, nil
 }
-
```
davidterpay, line 123: We inherit the default one so we don't need this since the check is the same.

abci/strategies/currencypair/default.go
```diff
@@ -114,9 +114,25 @@ func (s *DefaultCurrencyPairStrategy) GetDecodedPrice(
...
 	ctx sdk.Context,
 ) (uint64, error) {
-	return s.oracleKeeper.GetPrevBlockCPCounter(ctx)
+	current, err := s.oracleKeeper.GetNumCurrencyPairs(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	if mode := ctx.ExecMode(); mode == sdk.ExecModePrepareProposal || mode == sdk.ExecModeProcessProposal {
+		removed, err := s.oracleKeeper.GetNumRemovedCurrencyPairs(ctx)
```
aljo242, line 129:
```suggestion
		removed, err := s.oracleKeeper.GetNumRemovedCurrencyPairsInBlock(ctx)
```

I think let's be as pedantic as possible here since this is tricky enough to reason about

x/oracle/keeper/keeper.go
```diff
@@ -113,10 +113,19 @@ func NewKeeper(
 
 // RemoveCurrencyPair removes a given CurrencyPair from state, i.e. removes its nonce + QuotePrice from the module's store.
 func (k *Keeper) RemoveCurrencyPair(ctx sdk.Context, cp slinkytypes.CurrencyPair) error {
+	// check if the currency pair exists.
+	if !k.HasCurrencyPair(ctx, cp) {
+		return types.NewCurrencyPairNotExistError(cp)
```
aljo242, line 118: goated

x/oracle/keeper/keeper.go
```diff
@@ -113,10 +113,19 @@ func NewKeeper(
...
+	}
+
 	if err := k.currencyPairs.Remove(ctx, cp.String()); err != nil {
```
aljo242, line 121: this is just for knowledge sake - if this does not return an error if the key does not exist, what are the cases where this _does_ error?

abci/strategies/currencypair/default.go
g3ndo9kar9: Instead of implicitly casing on the exec-mode, I'd ideally like to expose to the caller whether they want the value of cps in state before the block was executed (specifically in `Prepare` / `Process`), or whether they want the # of cps in state (`Verify`)

g3ndo9kar9: This function has two diff behaviors AFAICT:
1. Get the # of cps in state before the block was executed (+ all cps added in the last block executed)
2. Get the current # of cps in state

davidterpay: that would a breaking change with the current dYdX integration so im in favor of keeping it as is.

g3ndo9kar9: This alr is a breaking change, no? Only the `GetPrevBlockCPCounter` is implemented in dydx

g3ndo9kar9:
So in this case, we're leaving `(numRemoved + numAddedCPs) * 33bz` bytes unaccounted for (if we're j checking the size of the vote-extension. What is the advantage of doing this, in comparison to checking per ID whether the cp reported for either existed in state or was removed (we originally proposed this)?

g3ndo9kar9 (commented):
left comment

aljo242:
> So in this case, we're leaving `(numRemoved + numAddedCPs) * 33bz` bytes unaccounted for (if we're j checking the size of the vote-extension. What is the advantage of doing this, in comparison to checking per ID whether the cp reported for either existed in state or was removed (we originally proposed this)?

The main one in my mind is that we are doing significantly less state reads this way. O(1) v O(n)

davidterpay:
@nivasan1 we originally had alignment that we would check only the size, you proposed the other idea but we decided to go with the simpler approach. I suggest we leave this as is and ask ottersec to take a deeper dive into this. More likely than not I can see the possibility of us moving into checking the IDs.

aljo242 (approved):
Approving for now but @davidterpay can you create an issue to clean this up / rename functions / doc this?

g3ndo9kar9 (approved):
lgtm

Eric-Warehime (approved):

PR #301, opened 2024-04-01
feat(defi): Implement raydium provider [BLO-1064]

g3ndo9kar9:
## In This PR
- I implement the `PriceFetcher` interface for the `raydium` provider


tests/integration/slinky_setup.go
```diff
@@ -362,7 +362,7 @@ func (s *SlinkyIntegrationSuite) AddCurrencyPairs(chain *cosmos.CosmosChain, aut
 				},
 			},
 			Paths: mmtypes.Paths{Paths: []mmtypes.Path{
-				mmtypes.Path{Operations: []mmtypes.Operation{
```
g3ndo9kar9, line 365: This file got added somehow for linting? @aljo242

tests/integration/slinky_slashing_setup.go
```diff
@@ -198,15 +198,18 @@ func UpdateNodePrices(node *cosmos.ChainNode, ticker mmtypes.Ticker, price int64
...
-			ticker.String(): {Paths: []mmtypes.Path{
-				{Operations: []mmtypes.Operation{
-					{
-						CurrencyPair: ticker.CurrencyPair,
-						Invert:       false,
-						Provider:     static.Name,
-					},
-				}}},
-			}},
+			ticker.String(): {
```
g3ndo9kar9, line 201: Same here

x/marketmap/types/ticker.go
```diff
@@ -57,7 +57,11 @@ func (t *Ticker) ValidateBasic() error {
 		return fmt.Errorf("metadata json field is longer than maximum length of %d", MaxMetadataJSONFieldLength)
 	}
 
-	return json.IsValid([]byte(t.Metadata_JSON))
+	if err := json.IsValid([]byte(t.Metadata_JSON)); err != nil {
+		return fmt.Errorf("invalid JSON metadata; %w", err)
+	}
```
g3ndo9kar9, line 62: This error was not informative enough IMO

tests/integration/slinky_setup.go
aljo242: just a formatting think i guess

x/marketmap/types/ticker.go
aljo242: https://github.com/skip-mev/slinky/pull/266/commits/5ca6a1e42f0606c5f223e79aed981f6eee429c24

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,252 @@
...
+	if err := market.ValidateBasic(); err != nil {
+		return nil, err
+	}
+
+	// check fields of config
+	if config.Name != Name {
+		return nil, fmt.Errorf("config.Name is not %s", Name)
+	}
+
+	if market.Name != Name {
```
aljo242: needs test case

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,252 @@
...
+	if err := config.ValidateBasic(); err != nil {
+		return nil, err
+	}
+
+	if err := market.ValidateBasic(); err != nil {
+		return nil, err
+	}
+
+	// check fields of config
+	if config.Name != Name {
```
aljo242: needs test case

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,252 @@
...
+	market oracletypes.ProviderMarketMap,
+	config config.APIConfig,
+	client SolanaJSONRPCClient,
+	logger *zap.Logger,
+) (*APIPriceFetcher, error) {
+	if err := config.ValidateBasic(); err != nil {
+		return nil, err
+	}
+
+	if err := market.ValidateBasic(); err != nil {
```
aljo242: needs test case

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,252 @@
...
+// for each currency-pair,
+//   - Query the raydium API base (coin) / quote (pc) token vault addresses
+//   - Normalize the token balances by 1e18
+//   - Calculate the price as quote / base, and scale by ticker.Decimals
+func (pf *APIPriceFetcher) FetchPrices(
+	ctx context.Context,
+	tickers []mmtypes.Ticker,
+) providertypes.GetResponse[mmtypes.Ticker, *big.Int] {
+	// get the acounts to query in order of the tickers given
+	accounts := make([]solana.PublicKey, len(tickers)*2)
```
aljo242: `len(tickers)*2` is used multiple times.

let's assign to a variable with a nice name

providers/apis/defi/raydium/provider.go
g3ndo9kar9: I'm testing this [here](https://github.com/skip-mev/slinky/pull/301/files#diff-e7c5b6ecb4619f7352f334ab7b4c7b5f046035f8a766924d5586dd80d329cd39R53) not sure why it doesn't show up in coverprof

providers/apis/defi/raydium/provider.go
g3ndo9kar9: tested [here](https://github.com/skip-mev/slinky/pull/301/files#diff-e7c5b6ecb4619f7352f334ab7b4c7b5f046035f8a766924d5586dd80d329cd39R79)

providers/apis/defi/raydium/provider.go
g3ndo9kar9: tested [here](https://github.com/skip-mev/slinky/pull/301/files#diff-e7c5b6ecb4619f7352f334ab7b4c7b5f046035f8a766924d5586dd80d329cd39R79)

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+		accounts []solana.PublicKey,
+		opts *rpc.GetMultipleAccountsOpts,
+	) (out *rpc.GetMultipleAccountsResult, err error)
+}
+
+// APIPriceFetcher is responsible for interacting with the solana API and querying information
+// about the price of a given currency pair.
+type APIPriceFetcher struct {
+	// market represents the ticker configurations for this provider.
+	market oracletypes.ProviderMarketMap
```
aljo242:
```suggestion
	// markets represents the ticker configurations for this provider.
	markets oracletypes.ProviderMarketMap
```

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+		metadata, err := unmarshalMetadataJSON(ticker.Metadata_JSON)
+		if err != nil {
+			return nil, fmt.Errorf("error unmarshalling metadata for ticker %s: %w", ticker.String(), err)
+		}
+
+		if err := metadata.ValidateBasic(); err != nil {
+			return nil, fmt.Errorf("metadata for ticker %s is invalid: %w", ticker.String(), err)
+		}
+
+		metadataPerTicker[ticker.String()] = metadata
```
aljo242: Is there a case where there is a duplicate `ticker` in OffchainMap?  Then there would be an overwrite into this map

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+				tickers,
+				providertypes.NewErrorWithCode(
+					NoRaydiumMetadataForTickerError(ticker.String()),
+					providertypes.ErrorUnknownPair,
+				),
+			)
+		}
+
+		accounts[i*2] = solana.MustPublicKeyFromBase58(metadata.BaseTokenVault.TokenVaultAddress)
+		accounts[i*2+1] = solana.MustPublicKeyFromBase58(metadata.QuoteTokenVault.TokenVaultAddress)
```
aljo242: Since these are always grouped together, could all of this array indexing logic be simplified by introducing a type:

```go
type PubKeyPair {
     BaseKey solana.PubKey
     QuoteKey solana.PubKey
}
```

?

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
+package raydium
```
davidterpay: small nit: can we rename the file to `fetcher.go` just so all of our provider implementations reference the underlying interfaces they are implementing?

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+
+	"github.com/skip-mev/slinky/oracle/config"
+	oracletypes "github.com/skip-mev/slinky/oracle/types"
+	"github.com/skip-mev/slinky/pkg/math"
+	"github.com/skip-mev/slinky/providers/base/api/handlers"
+	providertypes "github.com/skip-mev/slinky/providers/types"
+	mmtypes "github.com/skip-mev/slinky/x/marketmap/types"
+)
+
+var _ handlers.APIFetcher[mmtypes.Ticker, *big.Int] = (*APIPriceFetcher)(nil)
```
davidterpay:
```suggestion
var _ types.PriceAPIFetcher = (*APIPriceFetcher)(nil)
```
we already have a type alias for this.

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, fmt.Errorf("config for raydium is invalid: %w", err)
+	}
+
+	if err := market.ValidateBasic(); err != nil {
+		return nil, fmt.Errorf("market config for raydium is invalid: %w", err)
+	}
+
+	// check fields of config
+	if config.Name != Name {
+		return nil, fmt.Errorf("config.Name is not %s", Name)
```
davidterpay:
```suggestion
		return nil, fmt.Errorf("expected config name %s; got %s", Name, config.Name)
```
small nit

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	if market.Name != Name {
+		return nil, fmt.Errorf("market.Name is not %s", Name)
```
davidterpay: small nit: same here as above.

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+
+		metadataPerTicker[ticker.String()] = metadata
+	}
+
+	return &APIPriceFetcher{
+		market:            market,
+		config:            config,
+		client:            client,
+		metaDataPerTicker: metadataPerTicker,
+		logger:            logger.With(zap.String("provider", Name)),
```
davidterpay:
```suggestion
		logger:            logger.With(zap.String("api_price_fetcher", Name)),
```
We already add the provider tag to the zapper logger elsewhere so youll end up duplicating the same value.

davidterpay: Also better to maintain granularity on the struct this is actually referencing

providers/apis/defi/raydium/types.go
```diff
@@ -0,0 +1,74 @@
+package raydium
```
davidterpay, line 1: can you add a default api and market config for the provider?

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+
+		accounts[i*2] = solana.MustPublicKeyFromBase58(metadata.BaseTokenVault.TokenVaultAddress)
+		accounts[i*2+1] = solana.MustPublicKeyFromBase58(metadata.QuoteTokenVault.TokenVaultAddress)
+	}
+
+	// query the accounts
+	accountsResp, err := pf.client.GetMultipleAccountsWithOpts(ctx, accounts, &rpc.GetMultipleAccountsOpts{
+		Commitment: rpc.CommitmentFinalized,
+		// TODO(nikhil): Keep track of latest height queried as well?
+	})
```
davidterpay: Are we certain that the resp is sorted in the same order we make the request?

davidterpay: Is there a way for us to deterministically very this?

providers/apis/defi/raydium/provider.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// unmarshal the account data into a token account
+	var tokenAccount token.Account
+	if err := binary.NewBinDecoder(account.Data.GetBinary()).Decode(&tokenAccount); err != nil {
+		return nil, err
+	}
+
+	// get the token balance + scale by decimals
+	balance := new(big.Int).SetUint64(tokenAccount.Amount)
+	return balance.Mul(balance, new(big.Int).Exp(big.NewInt(10), big.NewInt(NormalizedTokenAmountExponent-int64(tokenDecimals)), nil)), nil
```
davidterpay: is there any chance that the token decimals is greater than the `NormalizedTokenAmountExponent` which is 18? Are you confident this is sufficiently high enough? i would look at some of the helper functions in the oracle agggreagtor funciton which has the functionality to scale up / down prices.

davidterpay (commented):
Left some comments, overall lgtm can approve once i get responses

providers/apis/defi/raydium/types.go
g3ndo9kar9: I'll make this in a separate PR?

providers/apis/defi/raydium/provider.go
g3ndo9kar9: The solana JSON-RPC server implementation orders the response in the order that accounts are requested. We don't get account addresses in response so we can't verify account data v account address

providers/apis/defi/raydium/provider.go
g3ndo9kar9: I'm p confident? None of the other serum tokens have more than this (at least those in raydium pools)

providers/apis/defi/raydium/provider.go
davidterpay: mmm

davidterpay: ok i would leave a godoc for that

providers/apis/defi/raydium/types.go
```diff
@@ -0,0 +1,74 @@
...
+	TokenVaultAddress string `json:"token_vault_address"`
+
+	// TokenDecimals is the number of decimals used for the token, we use this for
+	// normalizing the balance of tokens at the designated vault address
+	TokenDecimals uint64 `json:"token_decimals"`
+}
+
+// unmarshalMetadataJSON unmarshals the given metadata string into a TickerMetadata,
+// this method assumes that the metadata string is valid json, otherwise an error is returned.
+func unmarshalMetadataJSON(metadata string) (TickerMetadata, error) {
```
aljo242, line 56: Let's export and test

providers/apis/defi/raydium/types.go
```diff
@@ -0,0 +1,74 @@
...
+type TickerMetadata struct {
+	// BaseTokenVault is the metadata associated with the base token's token vault
+	BaseTokenVault AMMTokenVaultMetadata `json:"base_token_vault"`
+
+	// QuoteTokenVault is the metadata associated with the quote token's token vault
+	QuoteTokenVault AMMTokenVaultMetadata `json:"quote_token_vault"`
+}
+
+// ValidateBasic checks that the solana token vault addresses are valid.
+func (metadata TickerMetadata) ValidateBasic() error {
```
aljo242, line 29: Can we test this?

providers/apis/defi/raydium/price_fetcher.go
```diff
@@ -0,0 +1,256 @@
...
+		price := calculatePrice(baseTokenBalance, quoteTokenBalance, ticker.Decimals)
+
+		// return the price
+		resolved[ticker] = oracletypes.NewPriceResult(price, time.Now())
+	}
+
+	return oracletypes.NewPriceResponse(resolved, unresolved)
+}
+
+func getScaledTokenBalance(account *rpc.Account, tokenDecimals uint64) (*big.Int, error) {
```
aljo242, line 228: Can we test?

aljo242 (commented):
lgtm once we get some testing coverage

providers/apis/defi/raydium/price_fetcher.go
```diff
@@ -0,0 +1,256 @@
...
+	mmtypes "github.com/skip-mev/slinky/x/marketmap/types"
+)
+
+var _ oracletypes.PriceAPIFetcher = (*APIPriceFetcher)(nil)
+
+// SolanaJSONRPCClient is the expected interface for a solana JSON-RPC client according
+// to the APIPriceFetcher.
+//
+//go:generate mockery --name SolanaJSONRPCClient --output ./mocks/ --case underscore
+type SolanaJSONRPCClient interface {
```
Eric-Warehime, line 29: nit: I would probably put this interface in another package if it's supposed to be general across any solana defi provider we support.

providers/apis/defi/raydium/types.go
Eric-Warehime: This whole paradigm feels pretty fragile to me. I think we probably want to know how our arbitrary json objects are unmarhsalled way further up the stack and fail there instead of failing at the provider level. This feels like something that should be in the orchestrator--i.e. the request shouldn't even try to run unless we know we have the proper data to run it. I assume we'll want the same sort of behavior for any other provider using json metadata as well.

providers/apis/defi/raydium/price_fetcher.go
```diff
@@ -0,0 +1,256 @@
...
+				tickers,
+				providertypes.NewErrorWithCode(
+					NoRaydiumMetadataForTickerError(ticker.String()),
+					providertypes.ErrorUnknownPair,
+				),
+			)
+		}
+
+		accounts[i*2] = solana.MustPublicKeyFromBase58(metadata.BaseTokenVault.TokenVaultAddress)
+		accounts[i*2+1] = solana.MustPublicKeyFromBase58(metadata.QuoteTokenVault.TokenVaultAddress)
```
Eric-Warehime, line 151: Do we need to check for uniqueness here? We could be querying the same token vault a bunch if it's used in multiple pairs we care about right?

providers/apis/defi/raydium/price_fetcher.go
Eric-Warehime, line 1: None of this handles querying multiple RPC providers and choosing a result from those does it? It's fine if that's in a separate PR, just want to clarify.

g3ndo9kar9: see [this](https://github.com/skip-mev/slinky/pull/308)

providers/apis/defi/raydium/price_fetcher.go
g3ndo9kar9: That makes sense, I feel like we can do that when we implement the next provider? Or are you requesting a change to the dir structure of the `defi` directory?

providers/apis/defi/raydium/price_fetcher.go
g3ndo9kar9: Each token vault is an address that's separately derived from the pool that it's supposed to be querying, hence we expect all pool addresses to be unique

g3ndo9kar9: I can introduce a validity condition here as well?

providers/apis/defi/raydium/types.go
g3ndo9kar9: Agree w/ @Eric-Warehime, we discussed offline, since the change to introduce this validation is a larger change / scoped outside of this PR, I'll be handling this in a follow-on

providers/apis/defi/raydium/price_fetcher.go
Eric-Warehime: If every vault is already guaranteed unique then no need to check that.

providers/apis/defi/raydium/price_fetcher.go
Eric-Warehime: Yeah I suppose we can do it on the next solana provider.

Eric-Warehime (approved):

aljo242 (approved):
LGTM - let's make an issue to track further validation in a follow up PR after refactors

PR #302, opened 2024-04-01
feat(defi): Add Raydium logic to APIQueryHandler factory [BLO-1064]

g3ndo9kar9:
## In This PR
- I modify the `API`'s factory pattern to construct `APIFetcher`s in accordance with the provider being constructed


providers/factories/oracle/api.go
```diff
@@ -68,27 +74,41 @@ func APIQueryHandlerFactory(
...
+			requestHandler,
+			apiDataHandler,
+			metrics,
+			cfg.API,
+			logger,
+		)
 		if err != nil {
 			return nil, err
 		}
 	}
```
aljo242, line 113: Let's just move this logic before the `switch` statement.  Then this is just the initial value and can be overwritten if a custom one is provided.  We can remove the `apiPriceFetcher == nil` check that way as well

providers/base/api/handlers/api_query_handler.go
```diff
@@ -65,6 +65,21 @@ func NewAPIQueryHandler[K providertypes.ResponseKey, V providertypes.ResponseVal
...
+	priceFetcher, err := NewRestAPIFetcher(requestHandler, apiHandler, metrics, cfg, logger)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create price fetcher: %w", err)
+	}
+
+	return NewAPIQueryHandlerWithPriceFetcher(logger, cfg, priceFetcher, metrics)
+}
+
+// NewAPIQueryHandlerWithPriceFetcher creates a new APIQueryHandler with a custom price fetcher.
+func NewAPIQueryHandlerWithPriceFetcher[K providertypes.ResponseKey, V providertypes.ResponseValue](
```
aljo242, line 78: I would prefer we remove this new function and add variadic `Options` to `NewAPIQueryHandler`.  Then a `WithPriceFetcher` option can be passed in

providers/factories/oracle/api.go
g3ndo9kar9: we can't move the apiPriceFetcher creation to before the switch statement, as it depends on the apiDataHandler construction (this is what the switch is for) + we need that for the apiPriceFetcher

providers/base/api/handlers/api_query_handler.go
g3ndo9kar9: I don't think the optional `PriceFetcher` works in all cases? As there are two ways to approach this:
1. We pass in an un-needed `apiDataHandler`, in this case, if I'm making a defi provider (raydium / uniswap), we j pass in a nil `apiDataHandler`
2. We don't have the `apiDataHandler` as an argument, in this case, the `apiPriceFetcher` is not really optional?
Ultimately, the `QueryHandler` explicitly depends on the `PriceFetcher` so having it be an optional parameter does not seem like the correct move here?

I may be mis-understanding tho, why is the `Option` argument preferred?

davidterpay: I agree with alex that switching to options is eventually going to be better as we implement more implementations of the APIQueryHandler, but i think this is fine for the time being.

davidterpay: In this case the options would accept a api and request handler, the other option would only accept pricefetcher. several ways we could design it.

scripts/raydium/generate-config/main.go
```diff
@@ -0,0 +1,707 @@
+package main
```
davidterpay: Im fine having this script in the repo but we should have defaults in the implementation as well

Makefile
```diff
@@ -45,6 +45,15 @@ update-local-configs: build
 	@echo "Updating local config..."
 	@./build/slinky-config --oracle-config-path ${ORACLE_CONFIG_FILE} --market-config-path ${MARKET_CONFIG_FILE}
 
+generate-defi-configs:
+	@echo "Generating defi configs..."
+	@go run ./scripts/raydium/generate-config/main.go
+
+run-defi-providers: build generate-defi-configs
+	@echo "Running defi providers..."
+	@./build/slinky --oracle-config-path ./oracle.json --market-config-path ./market.json
```
davidterpay: why do we need a separate config for defi providers?

davidterpay (commented):
can we add defaults to the implementation?

davidterpay (approved):

aljo242 (approved):

scripts/raydium/generate-config/main.go
Eric-Warehime: Yeah I think it's probably necessary to have this somewhere. I hope it ends up being migrated to sauron pretty quickly though.

And I agree w/ David that we should either include a bunch of defaults in `/config/local/market.json` or have a separate config like `/config/raydium/market.json` where we can just run tests on raydium prices. Not sure what the ideal setup is there since we still need to integrate things into our dev/nightly tests more.

Eric-Warehime: Maybe there is some set of static values produced by this that we can include into `slinky-config` generation without having to pull in all of the node querying stuff?

oracle/types/oracle.go
```diff
@@ -62,6 +62,10 @@ type (
 	// returning the resolved and unresolved prices.
 	PriceAPIQueryHandler = apihandlers.APIQueryHandler[mmtypes.Ticker, *big.Int]
 
+	// APIPriceFetcher is a type alias for the API price fetcher. This is responsible
+	// for fetching the prices from the price provider using the API.
+	APIPriceFetcher = apihandlers.APIFetcher[mmtypes.Ticker, *big.Int]
```
Eric-Warehime: Duplicate of `PriceAPIFetcher` above I think.

Eric-Warehime (approved):

scripts/raydium/generate-config/main.go
g3ndo9kar9: I'm a bit confused on what the slinky-config script does? Seems like it just generates a test config for the oracle, and for dydx it generates the set of markets  + oracle json it needs to normal operation?

g3ndo9kar9: Since we aren't using this provider in prod, yet, I figured it was easiest to test w/ a separate config for raydium in particular. I can move to integrating this config into our own setup when I configure the dev slinky deployment

PR #307, opened 2024-04-04
feat: Add volatile market provider

Eric-Warehime:
Adds volatile market provider--mostly for use in dydx devnet, but can probably be useful in our testing as well.

Tested locally using static market config as well as using dydx localnet.


providers/volatile/price_test.go
```diff
@@ -0,0 +1,78 @@
...
+			name:          "test cosinePhase .25",
+			tp:            func() time.Time { return time.Unix(25*dailySeconds/100, 0) },
+			amplitude:     float64(0.95),
+			offset:        float64(100),
+			frequency:     float64(1),
+			expectedPrice: big.NewInt(5),
+		},
+		{
+			name:          "test cosinePhase .26",
+			tp:            func() time.Time { return time.Unix(dailySeconds/100, 0) },
```
aljo242:
```suggestion
			tp:            func() time.Time { return time.Unix(26 * dailySeconds/100, 0) },
```

Shouldn't it be this?

aljo242 (approved):

g3ndo9kar9 (approved):
lgtm, no websocket?

PR #308, opened 2024-04-04
feat(providers/api): multi-solana-client [BLO-1067]

g3ndo9kar9:
## In This PR
- I introduce an `Endpoints` field to the config, where we can place node-urls / authentication, etc
- I introduce a `MultiClient` to raydium to permit querying across multiple separated node-providers + aggregating repsonses from among them


oracle/config/api.go
```diff
@@ -51,7 +61,7 @@ func (c *APIConfig) ValidateBasic() error {
 		return fmt.Errorf("provider interval, timeout and reconnect timeout must be strictly positive")
 	}
 
-	if len(c.URL) == 0 {
+	if len(c.URL) == 0 && len(c.Endpoints) == 0 {
```
aljo242, line 64: the error message for this will now not be accurate.  Only mentions provider url

oracle/config/api_test.go
```diff
@@ -83,7 +83,7 @@ func TestAPIConfig(t *testing.T) {
 			expectedErr: true,
 		},
 		{
-			name: "bad config with no url",
+			name: "bad config with no url / endpoint",
```
aljo242, line 86: lets make a test for no url and a second one for no endpoints

oracle/config/api.go
```diff
@@ -33,10 +33,20 @@ type APIConfig struct {
 	// URL is the URL that is used to fetch data from the API.
 	URL string `json:"url"`
 
+	// Endpoints is a list of endpoints that the provider can query.
+	Endpoints []Endpoint `json:"endpoints"`
+
```
aljo242, line 38: can't this only be `Endpoints` and the minimum length just has to be 1?

can we remove URL?

Its unclear looking at the structure of the config what differentiates these two fields

providers/apis/defi/raydium/multi_client.go
```diff
@@ -0,0 +1,116 @@
...
+	for i := range endpoints {
+		client := rpc.New(endpoints[i].URL)
+		clients[i] = client
+	}
+	return NewMultiJSONRPCClient(clients, logger)
+}
+
+// GetMultipleAccountsWithOpts delegates the request to all underlying clients and applies a filter
+// to the responses.
+func (c *MultiJSONRPCClient) GetMultipleAccountsWithOpts(
```
aljo242, line 44: There is only one implementation of this function.

Can we just call it `GetMultipleAccounts`?

providers/apis/defi/raydium/multi_client.go
```diff
@@ -0,0 +1,116 @@
...
+	go func() {
+		select {
+		case <-ctx.Done():
+			c.logger.Error("context cancelled")
+		case <-channelForWaitGroup(&wg):
+		}
+		close(responsesCh)
+	}()
+
+	responses := make([]*rpc.GetMultipleAccountsResult, 0)
```
aljo242:
```suggestion
	responses := make([]*rpc.GetMultipleAccountsResult, len(responseCh))
```

we can preallocate this and not use `append` here

oracle/config/api.go
Eric-Warehime: I think the idea is to introduce the ability to specify multiple endpoints in a non-breaking way.

oracle/config/api_test.go
g3ndo9kar9: These all alr exist, see below

providers/apis/defi/raydium/multi_client.go
g3ndo9kar9: If there are failed responses / the channel is closed before all respond, we'll have nil entries?

g3ndo9kar9: I can set a capacity for the slice?

providers/apis/defi/raydium/multi_client.go
g3ndo9kar9: The interface is defined in accordance so that the default client generated stubs are an implementation, otherwise we'd have to make a wrapper / not use the getMultipleAccountsWithOpts method on the client

cmd/client/main.go
```diff
@@ -31,7 +31,7 @@ var (
 
 			// Set up a connection to the server.
 			url := fmt.Sprintf("%s:%s", host, port)
-			conn, err := grpc.Dial(url, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+			conn, err := grpc.NewClient(url, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
```
g3ndo9kar9, line 34: `grpc.Dial` is now deprecated in favor of `NewClient`

cmd/slinky-config/main.go
```diff
@@ -323,11 +323,11 @@ func init() {
 		"whether or not to enable raydium support",
 	)
 	rootCmd.Flags().StringVarP(
-		&solanaNodeURL,
+		&solanaNodeURLs,
 		"solana-node-endpoint",
 		"",
 		"",
-		"The HTTP endpoint of the solana node endpoint the raydium provider will be configured to use",
+		"The HTTP endpoints of the solana node endpoint the raydium provider will be configured to use, notice, if multiple are given they must be comma delimited",
```
aljo242:
```suggestion
		"The HTTP endpoints of the solana node endpoint the raydium provider will be configured to use. If multiple are given they must be comma delimited",
```

aljo242 (approved):

cmd/slinky-config/main.go
```diff
@@ -375,7 +375,13 @@ func createOracleConfig() error {
...
 	if raydiumEnabled {
 		cfg := raydium.DefaultAPIConfig
-		cfg.URL = solanaNodeURL
+		nodes := strings.Split(solanaNodeURLs, ",")
+
+		for _, node := range nodes {
+			cfg.Endpoints = append(cfg.Endpoints, config.Endpoint{
+				URL: node,
+			})
+		}
```
Eric-Warehime: The solanaNodeURLs flag should just be a [StringSliceVarP](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringSliceVarP). It eliminates all of this code.

providers/apis/defi/raydium/multi_client.go
```diff
@@ -0,0 +1,116 @@
...
+			c.logger.Error("context cancelled")
+		case <-channelForWaitGroup(&wg):
+		}
+		close(responsesCh)
+	}()
+
+	responses := make([]*rpc.GetMultipleAccountsResult, 0, len(c.clients))
+	for resp := range responsesCh {
+		responses = append(responses, resp)
+	}
```
Eric-Warehime:
```suggestion
	responses := make([]*rpc.GetMultipleAccountsResult, 0, len(c.clients))

	// spawn a goroutine for each client to fetch the accounts
	var wg sync.WaitGroup
	wg.Add(len(c.clients))

	for i := range c.clients {
		go func(client SolanaJSONRPCClient) {
			defer wg.Done()
			resp, err := client.GetMultipleAccountsWithOpts(ctx, accounts, opts)
			if err != nil {
				c.logger.Error("failed to fetch accounts", zap.Error(err))
				return
			}
			responses[i] = resp
		}(c.clients[i])
	}
	wg.Wait()
```

Eric-Warehime: I'm struggling to see why we need anything more than parallelized fetches here. Anything that errors out will properly have no results, and we block on all the responses returning no matter what.

g3ndo9kar9: This logic defers to context closure over all clients finishing their outbound requests. Agreed that the behavior is the same if the underlying client defers to context closure

Eric-Warehime (approved):

aljo242 (approved):

PR #311, opened 2024-04-04
feat: UniswapV3 Provider

davidterpay:
closes: BLO-1054

Uniswap V3 Provider implementation

Note: around ~2800 LOC from this PR is from code-generation, much more is testing. I recommend reading over the readme before reviewing


providers/apis/uniswapv3/utils.go
```diff
@@ -0,0 +1,67 @@
...
+// DefaultAPIConfig is the default configuration for the Uniswap API.
+var DefaultAPIConfig = config.APIConfig{
+	Name:             Name,
+	Atomic:           true,
+	Enabled:          true,
+	Timeout:          1000 * time.Millisecond,
+	Interval:         2000 * time.Millisecond,
+	ReconnectTimeout: 2000 * time.Millisecond,
+	MaxQueries:       1,
+	URL:              "https://eth.public-rpc.com/",
```
davidterpay: This RPC is probably fine to have displayed on our main, can be a good default for people.

pkg/math/math_test.go
```diff
@@ -54,6 +54,45 @@ func TestMin(t *testing.T) {
 	}
 }
 
+func TestAbs(t *testing.T) {
+	t.Parallel()
```
aljo242: no need for this

pkg/math/math.go
```diff
@@ -19,6 +19,14 @@ func Min[V int | int64 | uint64 | int32 | uint32](vals ...V) V {
 	return minimum
 }
 
+// Abs returns the absolute value of a given number.
+func Abs[V int | int32 | int64](val V) V {
```
aljo242: we should use this predefined type constraint 

https://pkg.go.dev/golang.org/x/exp/constraints#Signed

pkg/math/test_utils.go
```diff
@@ -0,0 +1,51 @@
+package math
```
aljo242: nit: rename to `testutils.go`

providers/apis/uniswapv3/fetcher.go
```diff
@@ -0,0 +1,269 @@
...
+	if err := api.ValidateBasic(); err != nil {
+		return nil, fmt.Errorf("invalid api config: %w", err)
+	}
+
+	abi, err := uniswappool.UniswapMetaData.GetAbi()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get uniswap abi: %w", err)
+	}
+
+	payload, err := abi.Pack("slot0")
```
aljo242: can we make "slot0" constant and doc its usage?

providers/apis/uniswapv3/fetcher.go
```diff
@@ -0,0 +1,269 @@
...
+// and validate the pool config which contains all required information to query the EVM.
+func (u *PriceFetcher) GetPool(
+	ticker mmtypes.Ticker,
+) (PoolConfig, error) {
+	if pool, ok := u.poolCache[ticker]; ok {
+		return pool, nil
+	}
+
+	var cfg PoolConfig
+	if err := json.Unmarshal([]byte(ticker.Metadata_JSON), &cfg); err != nil {
```
aljo242: like raydium - this is fine for now, but lets make a tracking issue to have this unpacking / validation occur earlier

providers/apis/uniswapv3/fetcher_test.go
```diff
@@ -0,0 +1,256 @@
...
+			name: "weth/usdc mainnet result",
+			tickers: []mmtypes.Ticker{
+				weth_usdc_ticker,
+			},
+			client: func() uniswapv3.EVMClient {
+				batchErrors := []error{
+					nil,
+				}
+				responses := []string{
+					"0x00000000000000000000000000000000000043dd3b966e761000000000000000000000000000000000000000000000000000000000000000000000000002fabf000000000000000000000000000000000000000000000000000000000000057900000000000000000000000000000000000000000000000000000000000005a000000000000000000000000000000000000000000000000000000000000005a000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001",
```
aljo242: beautiful

providers/apis/uniswapv3/utils.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if pc.QuoteDecimals <= 0 {
+		return fmt.Errorf("quote decimals must be positive")
+	}
+
+	return nil
+}
+
+// ToJSON converts the pool configuration to JSON.
+func (pc *PoolConfig) ToJSON() string {
```
aljo242: nit: rename to `MustToJSON` since this can `panic`  

Might also just be better to return error here

davidterpay: it'll be a bit annoying having to case on everything if it does return an error, ill rename to MustToJSON tho

providers/apis/uniswapv3/fetcher.go
davidterpay: I ref primer for people that want to understand slot0 better.

davidterpay: I'm fine also making this a constant, but given that we need the ABI either way i thought it would just be easier to have the payload be generated on start. no strong preference here

aljo242: no, i agree with that. 

I just mean the string literal "slot0" as it is used a few times

aljo242:
we got conflicts :(

providers/apis/uniswapv3/fetcher.go
davidterpay: ohhh

davidterpay: Yea lets do that

davidterpay:
btw our lint is marking files i did not change with errors

aljo242 (approved):

README.md
```diff
@@ -98,21 +98,3 @@ To test these numbers yourself, spin up the the oracle server following the inst
...
-
-Then, to run
-
-```bash
-$ make run-oracle-server
-```
-
-## Future Work
-
-The oracle side car is a combination of the oracle and provider packages. This is being moved to a [separate repository](https://github.com/skip-mev/slinky-sidecar).
```
Eric-Warehime, line 113: Definitely want this removed 👍 wasn't aware this was in here.

README.md
```diff
@@ -98,21 +98,3 @@ To test these numbers yourself, spin up the the oracle server following the inst
 * [Data Provider Queries](./providers/base/metrics/README.md#usage): Provides general insight into how often price feeds are updated by status (success/failure), provider (binance, coinbase, etc.), price feed (BTC/USD, ETH/USD), and provider type (api/websocket).
 * [Websocket Performance Queries](./providers/base/websocket/metrics/README.md#usage): Provides insight into how often websocket providers are successfully updating their data. This is a combination of metrics related to the underlying connection as well as the data handler which is responsible for processing the data received from the Websocket connection.
 * [API Performance Queries](./providers/base/api/metrics/README.md#usage): Provides insight into how often API providers are successfully updating their data.
-
-## DEFI
```
Eric-Warehime, line 102: You need to sync w/ @nivasan1 because I think he's just added all of this and you're removing it now.

go.mod
```diff
@@ -25,8 +25,9 @@ require (
 	github.com/cosmos/cosmos-sdk v0.50.5
 	github.com/cosmos/gogogateway v1.2.0
 	github.com/cosmos/gogoproto v1.4.12
-	github.com/gagliardetto/binary v0.7.7
-	github.com/gagliardetto/solana-go v1.9.3
+	github.com/ethereum/go-ethereum v1.13.14
+	github.com/gagliardetto/binary v0.8.0
+	github.com/gagliardetto/solana-go v1.10.0
```
Eric-Warehime, line 30: @nivasan1 do we have sufficient coverage to convince us that this is ok based on CI passing?

providers/apis/defi/uniswapv3/client.go
```diff
@@ -0,0 +1,41 @@
...
+	// BatchCallContext is a batch call to an EVM.
+	BatchCallContext(ctx context.Context, calls []rpc.BatchElem) error
+}
+
+var _ EVMClient = (*GoEthereumClientImpl)(nil)
+
+// GoEthereumClient is a go-ethereum client implementation using the go-ethereum RPC
+// library.
+type GoEthereumClientImpl struct {
+	client *rpc.Client
```
g3ndo9kar9, line 22: Why do we need to wrap?

providers/apis/defi/uniswapv3/utils.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if pc.QuoteDecimals <= 0 {
+		return fmt.Errorf("quote decimals must be positive")
+	}
+
+	return nil
+}
+
+// MustToJSON converts the pool configuration to JSON.
+func (pc *PoolConfig) MustToJSON() string {
```
g3ndo9kar9, line 54: Where is this used?

providers/apis/defi/uniswapv3/fetcher.go
```diff
@@ -0,0 +1,269 @@
...
+		)
+
+		return types.NewPriceResponseWithErr(
+			tickers,
+			providertypes.NewErrorWithCode(err, providertypes.ErrorAPIGeneral),
+		)
+	}
+
+	// Parse the result from the batch call for each ticker.
+	for i, ticker := range tickers {
```
g3ndo9kar9, line 171: Is there a way to guarantee the order of responses from batch-json RPC requests?

g3ndo9kar9 (commented):
2 small comments. My follow up thoughts:
1. Do we need to be able to correlate requests among multiple rpc-providers (like solana)
2. As needed for raydium, we'll need to have some way to verify that the ProviderMetadata per ticker unmarshals correctly into the expected json-data structure. Ideally, we'd be able to do this on market-map validation

pkg/math/testutils.go
```diff
@@ -0,0 +1,51 @@
...
+	if expected.Cmp(zero) == 0 {
+		require.Equal(t, zero, actual)
+		return
+	}
+
+	var diff *big.Float
+	if expected.Cmp(actual) > 0 {
+		diff = new(big.Float).Sub(new(big.Float).SetInt(expected), new(big.Float).SetInt(actual))
+	} else {
+		diff = new(big.Float).Sub(new(big.Float).SetInt(actual), new(big.Float).SetInt(expected))
```
Eric-Warehime, line 29: Can't you just use the `Abs` function here?

pkg/math/oracle/median_test.go
```diff
@@ -149,7 +150,7 @@ func TestAggregateData(t *testing.T) {
 			result := m.DataAggregator.GetAggregatedData()
 			require.Equal(t, len(tc.expectedPrices), len(result))
 			for ticker, price := range result {
-				verifyPrice(t, tc.expectedPrices[ticker], price)
+				math.VerifyPrice(t, tc.expectedPrices[ticker], price, acceptableDelta)
```
Eric-Warehime, line 153: Why did these become non-deterministic now?

davidterpay: they aren't non-deterministic, they just put a bound on the amount of precision we expect in the resulting calculation. This only applies for big.Ints, for big.Floats we can directly specify precision so much of this might be deprecated with the changes that are following this.

pkg/math/testutils.go
davidterpay: I don't fully remember why i didn't use `Abs` but i do remember not using it for a specific reason. This code might change as we convert to big.Floats so I'd prefer leaving it as is for now.

providers/apis/defi/uniswapv3/fetcher.go
davidterpay: the client should return things in the same order, but we can probably resort just for sanity before using the responses.

providers/apis/defi/uniswapv3/fetcher.go
```diff
@@ -0,0 +1,269 @@
...
+	// payload is the packed slot0 call to the pool contract. Since the slot0 payload is the same
+	// for all pools, we can reuse this payload for all pools.
+	payload []byte
+	// poolCache is a cache of the tickers to pool configs. This is used to avoid unmarshalling
+	// the metadata for each ticker.
+	poolCache map[mmtypes.Ticker]PoolConfig
+}
+
+// NewPriceFetcher returns a new Uniswap V3 price fetcher.
+func NewPriceFetcher(
```
Eric-Warehime, line 55: nit: Untested.

Eric-Warehime (approved):

PR #315, opened 2024-04-05
feat: Compatibility between the voteweighted package and ICS consumer keeper

Eric-Warehime:
Adds compat types so that ICS chains can use slinky.

It uses validator power instead of bonded tokens in the stake-weighted median computation.

It uses the most recent version of ics v5.0.0-alpha1.


NeverHappened (commented):
Seems nice! Two requests:
- there is no way to create CCVConsumerCompatKeeper now, because ccvKeeper is a private field. Can you add a constructor to it please?
- can you also implement another ValidatorStore as well for CCVConsumerCompatKeeper? (it's used in another place)
```
type ValidatorStore interface {
	GetPubKeyByConsAddr(context.Context, sdk.ConsAddress) (cmtprotocrypto.PublicKey, error)
}
```

I implemented it like this (seems to be working):
```
func (c ConsumerValidatorStore) GetPubKeyByConsAddr(ctx context.Context, consAddr sdk.ConsAddress) (cmtprotocrypto.PublicKey, error) {
	sdkCtx := sdk.UnwrapSDKContext(ctx)
	val, found := c.k.GetCCValidator(sdkCtx, consAddr)
	if !found {
		// TODO
		return cmtprotocrypto.PublicKey{}, fmt.Errorf("not found CCValidator for address = TODO: ")
	}

	consPubKey, err := val.ConsPubKey()
	if err != nil {
		// TODO
		return cmtprotocrypto.PublicKey{}, fmt.Errorf("TODO")
	}
	tmPubKey, err := cryptocodec.ToCmtProtoPublicKey(consPubKey)
	if err != nil {
		// TODO
		return cmtprotocrypto.PublicKey{}, err
	}

	return tmPubKey, nil
}
```

pkg/math/voteweighted/ccv_compat.go
```diff
@@ -0,0 +1,59 @@
...
+	if !found {
+		return nil, fmt.Errorf("could not find validator %s", addr.String())
+	}
+	return CCVCompat{ccv: ccv}, nil
+}
+
+// TotalBondedTokens iterates through all CCVs and returns the sum of all validator power.
+func (c CCVConsumerCompatKeeper) TotalBondedTokens(ctx context.Context) (math.Int, error) {
+	total := math.NewInt(0)
+	sdkCtx, ok := ctx.(sdk.Context)
```
NeverHappened: how about using `	sdkCtx := sdk.UnwrapSDKContext(ctx)`?

g3ndo9kar9 (approved):
lgtm

aljo242 (approved):

PR #316, opened 2024-04-05
fix: compute gov addr after `sdk.Config()` changed

beer-1:
`authtypes.NewModuleAddress` internally use `sdk.Config()` to decide the bech32 prefix of the acc address. 

In current implementation, the function is called before the config modification. so it is using `cosmos` prefix to compute gov module address.

This lead `sdk.AccAddress{}.String()` always compute the address with `cosmos` bech32 prefix due to cache logic inside `sdk.AccAddress`. Also this leads gov module address stored with `cosmos1` prefixed string address in any chain.


aljo242 (approved):

g3ndo9kar9 (approved):
lgtm

Eric-Warehime (approved):

PR #321, opened 2024-04-05
chore: error log instead of debug on failed data fetch in base provider

aljo242:


davidterpay:
Can you run the side-car and make sure this isn't propagating like crazy in the side-car

aljo242:
> Can you run the side-car and make sure this isn't propagating like crazy in the side-car

Yeah we gucci

Eric-Warehime (approved):

davidterpay (approved):

PR #322, opened 2024-04-05
refactor: Provider Ticker Interface (PR1)

davidterpay:
NOTE:
Most of the changes are updating the market and oracle configs.


oracle/types/oracle.go
```diff
@@ -14,7 +14,6 @@ import (
 	wsmetrics "github.com/skip-mev/slinky/providers/base/websocket/metrics"
 	providertypes "github.com/skip-mev/slinky/providers/types"
 	"github.com/skip-mev/slinky/providers/types/factory"
-	mmtypes "github.com/skip-mev/slinky/x/marketmap/types"
 )
```
davidterpay, line 18: Note, every single alias that is relevant to a provider will now utilize the ProviderTicker interface as oppose to mmtypes.Ticker. The aggregator now maps (provider, off-chain-ticker) -> price. This will make it much easier to index prices from the aggregator with a specified market map.

aljo242:
run `make format` plz

davidterpay:
@aljo242 is that fine to run with broken changes?

oracle/types/market.go
```diff
@@ -1,160 +1,94 @@
...
+	var (
+		// Track all of the tickers that the provider will be providing data for.
+		providerTickers = make([]ProviderTicker, 0)
+		// Maintain a set of off-chain tickers that have been seen to avoid duplicates.
+		seenOffChainTickers = make(map[string]struct{})
+	)
+
+	// Iterate through every single market and its provider configurations to find the
+	// provider configurations that match the provider name.
+	for _, market := range marketMap.Markets {
```
aljo242, line 30: we can probably optimize this a bit in the future (right now we are doing O(n2) search and can probably make it O(n) with some caching or awareness between providers when constructing.)

not crucial now

oracle/types/market.go
```diff
@@ -1,160 +1,94 @@
...
 }
 
-// GetTickers returns the tickers from the provider market map.
-func (pmm *ProviderMarketMap) GetTickers() []mmtypes.Ticker {
-	tickers := make([]mmtypes.Ticker, 0, len(pmm.TickerConfigs))
-	for ticker := range pmm.TickerConfigs {
-		tickers = append(tickers, ticker)
-	}
-	return tickers
+// SimpleProviderTicker is a simple representation of a provider ticker. Provider's
```
aljo242:
```suggestion
// SimpleProviderTicker is a simple representation of a provider ticker. Providers
```

oracle/types/market.go
```diff
@@ -1,160 +1,94 @@
...
+// can utilize this to easily configure custom tickers.
+type SimpleProviderTicker struct {
```
aljo242: Would prefer calling this "Base"

oracle/types/market.go
```diff
@@ -1,160 +1,94 @@
...
+type TickersToProviderTickers map[mmtypes.Ticker]SimpleProviderTicker
 
-	// Read the entire file at the given path
-	data, err := os.ReadFile(path)
-	if err != nil {
-		return config, fmt.Errorf("error reading config file: %w", err)
-	}
+// MustToProviderTickers converts the map to a list of provider tickers.
+func (tpt *TickersToProviderTickers) MustToProviderTickers() []ProviderTicker {
+	var providerTickers []ProviderTicker
```
aljo242: let's preallocate like the linter suggest

oracle/types/oracle.go
```diff
@@ -24,113 +23,114 @@ type (
...
 	// This is responsible for parsing web socket messages and returning the resolved
 	// and unresolved prices.
-	PriceWebSocketDataHandler = wshandlers.WebSocketDataHandler[mmtypes.Ticker, *big.Int]
+	PriceWebSocketDataHandler = wshandlers.WebSocketDataHandler[ProviderTicker, *big.Int]
 
 	// PriceWebSocketQueryHandler is a type alias for the price web socket query handler.
 	// This is responsible for building the web socket query for the price provider and
 	// returning the resolved and unresolved prices.
-	PriceWebSocketQueryHandler = wshandlers.WebSocketQueryHandler[mmtypes.Ticker, *big.Int]
+	PriceWebSocketQueryHandler = wshandlers.WebSocketQueryHandler[ProviderTicker, *big.Int]
```
aljo242: ty type aliases

oracle/types/market.go
aljo242: why do we have SimpleProviderTicker and DefaultPRoviderTIcker?

oracle/types/provider.go
```diff
@@ -0,0 +1,111 @@
...
+	// Provider's that utilize this implementation should be able to easily configure
+	// custom json data for their tickers.
+	DefaultProviderTicker struct {
+		provider string
+		offChain string
+		decimals uint64
+		json     string
+	}
+
+	// ProviderTickers is a type alias for a list of provider tickers.
```
Eric-Warehime: This is not a list.

x/mm2/types/utils.go
Eric-Warehime, line 1: This is a weird thing to include in this PR. Not sure how we didn't already have this function also.

oracle/types/oracle.go
Eric-Warehime: I think the following is true...it's been hard to reason about things because of the rewrites going on. Let me know if this makes sense or is completely wrong.

This means we now have 2 uniqueness constraints right? `mmtypes.Ticker`--this is what we would call a "canonical ticker" or something like that which is `base/quote`. This has to be unique across all Markets and that is enforced in the keeper.

And now we have `off-chain-ticker` which is similar, but only has to be unique per Provider. That's not really enforced anywhere. It's checked in `ProviderTickersFromMarketMap`. If we hit an offchain ticker we've already seen, we ignore it. Right now the only thing that could be different between them is the json metadata so it's probably not a big deal, but it would result in indeterminism based on iteration order.

oracle/types/market_test.go
```diff
@@ -5,229 +5,216 @@ import (
...
+						ProviderConfigs: []mmtypes.ProviderConfig{
+							{
+								Name:           "test",
+								OffChainTicker: "ETH/USDT",
+								Invert:         true,
+								NormalizeByPair: &pkgtypes.CurrencyPair{
+									Base:  "ETH",
+									Quote: "USD",
+								},
+								Metadata_JSON: "{}",
```
Eric-Warehime, line 114: If this field was different from the metadata json for the other ProviderConfig, what is the expected behavior?

oracle/types/oracle.go
davidterpay: Yes, there should be a unique-ness check based on the off-chain-ticker and the json metadata. Our assumption atm is that In no case should the same provider have the same off-chain-ticker but have different json meta-data.

davidterpay: we should probably add that as a check in the module.

oracle/types/market.go
davidterpay: yea im gonna remove the other

davidterpay: removed

oracle/types/market.go
```diff
@@ -1,160 +1,66 @@
...
+	return providerTickers, nil
 }
 
-// ReadMarketConfigFromFile reads a market map configuration from a file at the given path.
-func ReadMarketConfigFromFile(path string) (mmtypes.MarketMap, error) {
-	// Initialize the struct to hold the configuration
-	var config mmtypes.MarketMap
+// TickersToProviderTickers is a map of tickers to provider tickers. This should be
+// utilized by providers to configure the tickers they will be providing data for.
+type TickersToProviderTickers map[mmtypes.Ticker]DefaultProviderTicker
```
aljo242: I think it probably makes more sense for this to be.  We can just use the interface

```suggestion
type TickersToProviderTickers map[mmtypes.Ticker]ProviderTicker
```

davidterpay: I need this to be the concrete type since I am going to be explicitly assigning DefaultProviderTicker. I would check out the next PR.

oracle/types/market.go
```diff
@@ -1,160 +1,66 @@
...
-			return fmt.Errorf("off-chain ticker %s not found in off-chain map", config.OffChainTicker)
-		}
-
-		if t != ticker {
-			return fmt.Errorf("off-chain ticker %s does not match on-chain ticker %s", config.OffChainTicker, ticker)
+			providerTicker := NewProviderTicker(
+				cfg.Name,
+				cfg.OffChainTicker,
+				cfg.Metadata_JSON,
+				DefaultTickerDecimals,
```
g3ndo9kar9: Why are we using a default TickerDecimals here? Should this even be relevant to providers if we're moving to `big.Float` arithmetic in providers?

oracle/types/market.go
```diff
@@ -1,160 +1,66 @@
...
+	)
+
+	// Iterate through every single market and its provider configurations to find the
+	// provider configurations that match the provider name.
+	for _, market := range marketMap.Markets {
+		for _, cfg := range market.ProviderConfigs {
+			if cfg.Name != name {
+				continue
+			}
+			if _, ok := seenOffChainTickers[cfg.OffChainTicker]; ok {
```
g3ndo9kar9, line 35: This feels like a validation that should be performed in the `ValidateBasic` for `ProviderConfig`?

g3ndo9kar9: Also should we not be failing here if we see two separate offchain tickers in a `ProviderConfig` as we're iterating over a map it's not clear which we choose?

oracle/types/market.go
g3ndo9kar9: Why do we need this at all? Shouldn't each provider just be able to take the `[]ProviderTicker`? My intuition is that, ideally, there's no state per ticker in each provider. All info per ticker a provider needs should be accessible from the `ProviderTicker` iface. Reason being, we'll be changing metadata for each ticker (suppose we change decimals or `Metadata_JSON` for a ticker). it's much simpler for this to just work (pass `ProviderTickers`), w/o awareness of each provider internal (having to selectively update providers' ticker state).

oracle/types/oracle.go
```diff
@@ -24,113 +23,114 @@ type (
...
 
 	// PriceResponse is a type alias for the price response. A price response is
 	// composed of a map of resolved prices and a map of unresolved prices. Resolved
 	// prices are the prices that were successfully fetched from the API, while
 	// unresolved prices are the prices that were not successfully fetched from the API.
-	PriceResponse = providertypes.GetResponse[mmtypes.Ticker, *big.Int]
+	PriceResponse = providertypes.GetResponse[ProviderTicker, *big.Int]
```
g3ndo9kar9: Is it safe to use an interface as a reference into a map (ifaces are reference types). Can't we just use the OffChainTicker string as the index into `ResolvedPrices` for each provider?

oracle/types/oracle.go
```diff
@@ -24,113 +23,114 @@ type (
...
-	UnResolvedPrices = map[mmtypes.Ticker]providertypes.UnresolvedResult
+	UnResolvedPrices = map[ProviderTicker]providertypes.UnresolvedResult
 
 	// TickerPrices is a type alias for the map of prices. This is a map of tickers i.e.
 	// BTC/USD, ETH/USD, etc. to their respective prices.
-	TickerPrices = map[mmtypes.Ticker]*big.Int
+	TickerPrices = map[ProviderTicker]*big.Int
+
+	// AggregatorPrices is a type alias for a map of off-chain ticker to the price.
+	AggregatorPrices = map[string]*big.Int
```
g3ndo9kar9: What's the diff between this and `ResolvedPrices

oracle/types/provider.go
```diff
@@ -0,0 +1,111 @@
...
+	DefaultProviderTicker struct {
+		provider string
+		offChain string
+		decimals uint64
+		json     string
+	}
+
+	// ProviderTickers is helper struct to manage a list of provider tickers.
+	ProviderTickers struct {
+		cache map[string]ProviderTicker
```
g3ndo9kar9, line 31: nit: what's the index here?

oracle/types/provider.go
```diff
@@ -0,0 +1,111 @@
...
+		"provider: %s, off-chain-ticker: %s, decimals: %d, json: %s",
+		t.provider,
+		t.offChain,
+		t.decimals,
+		t.json,
+	)
+}
+
+// NewProviderTickers returns a new list of provider tickers.
+func NewProviderTickers(tickers ...ProviderTicker) ProviderTickers {
```
g3ndo9kar9, line 61: Where are we using this?

g3ndo9kar9 (commented):
left comments

oracle/types/provider.go
davidterpay: check out the later PRs, we use it as a helper function to get more or less the same functionality we had with the ProviderMarketMap. just helps maintain internal state of things we are going to need a good amount (like off-chain-ticker to provider ticker mappings).

oracle/types/provider.go
davidterpay: off-chain-ticker

oracle/types/market.go
davidterpay: each provider does take in `[]ProviderTicker`. check out the follow up PRs.

davidterpay: the handlers are all essentially stateless now, except for the use of `types.ProviderTickers` which providers some basic utility functions.

oracle/types/market.go
davidterpay: hmm, yea i guess scaling up the values without a notion of the necessary precision before hitting aggregation probably just adds overhead.

davidterpay: good point

oracle/types/market.go
davidterpay: Yea i brought this up in slack.

davidterpay: @aljo242

oracle/types/oracle.go
davidterpay: I would check out the follow up PRs, but using only the offchain ticker as a primary key would not give us the amount of info we need to determine what we are parsing when we get a response.

oracle/types/oracle.go
davidterpay: each provider is responsible for returning prices per ProviderTicker, the aggregator is only aware of [providerName, off-chain-ticker] -> price. So when we have a market map we know how to directly consume each of the prices. This change will be more understandable once i get around to updating the main oracle.

oracle/constants/currency_pairs.go
```diff
@@ -0,0 +1,132 @@
+package constants
```
davidterpay, line 1: Per Nikhil's point, we should be able to define cannonical representations of the feeds without enforcing any additional criteria (min provider count, decimals, etc). Those can be determined adhoc instead of having to explicitly define that when we might want (dYdX decimals, initia decimals)

providers/factories/oracle/factory.go
```diff
@@ -1,124 +0,0 @@
-package oracle
```
davidterpay, line 1: Dead code

providers/volatile/api_handler.go
```diff
@@ -35,48 +32,28 @@ var DefaultAPIConfig = config.APIConfig{
...
-		if _, ok := v.tickers[ticker]; ok {
-			resolved[ticker] = types.NewPriceResult(volatilePrice, time.Now())
-		} else {
-			err := fmt.Errorf("failed to resolve ticker %s", ticker)
-			unresolved[ticker] = providertypes.UnresolvedResult{
-				ErrorWithCode: providertypes.NewErrorWithCode(err, providertypes.ErrorUnknownPair),
-			}
-		}
+		resolved[ticker] = types.NewPriceResult(volatilePrice, time.Now())
 	}
```
davidterpay: This is returning the same price for every ticker, which was the behavior beforehand but im not sure if its intended. gonna leave this as is

cmd/slinky-config/main.go
```diff
@@ -565,17 +540,12 @@ func addRaydiumMarkets(providerToMarkets map[string]map[mmtypes.Ticker]mmtypes.P
 	}
 
 	// add the raydium markets to the provider to markets map
-	providerToMarkets[raydium.Name] = make(map[mmtypes.Ticker]mmtypes.ProviderConfig)
+	providerToMarkets[raydium.Name] = make(types.CurrencyPairsToProviderTickers)
```
davidterpay, line 541: @nivasan1 can you make sure this still functions as desired.

service/servers/oracle/helpers.go
```diff
@@ -4,11 +4,12 @@ import (
...
 )
 
-func ToReqPrices(prices types.TickerPrices) map[string]string {
+func ToReqPrices(prices types.AggregatorPrices) map[string]string {
 	reqPrices := make(map[string]string, len(prices))
 
 	for cp, price := range prices {
-		reqPrices[cp.String()] = price.String()
+		intPrice, _ := price.Int(nil)
+		reqPrices[cp] = intPrice.String()
```
davidterpay, line 12: Truncating at the latest possible point in the lifecycle of a price.

x/mm2/types/market.go
```diff
@@ -57,10 +57,11 @@ func (m *Market) ValidateBasic() error {
...
 
 		// check for duplicate providers
-		if _, seen := seenProviders[providerConfig.Name]; seen {
-			return fmt.Errorf("duplicate provider found: %s", providerConfig.Name)
+		key := providerConfig.Name + providerConfig.OffChainTicker
+		if _, seen := seenProviders[key]; seen {
+			return fmt.Errorf("duplicate (provider, off-chain-ticker) found: %s", key)
 		}
-		seenProviders[providerConfig.Name] = struct{}{}
+		seenProviders[key] = struct{}{}
```
davidterpay: A given provider might show up twice in a given market, for example BTC-USD and BTC-USDT for coinbase.

davidterpay: We should have a separate check that the off-chain-ticker has a unique json field just so that aren't two separate (off-chain, json) fields.

oracle/oracle.go
```diff
@@ -227,7 +222,7 @@ func (o *OracleImpl) fetchPrices(provider types.PriceProviderI) {
 			zap.String("price", result.Value.String()),
 			zap.Duration("diff", diff),
 		)
-		timeFilteredPrices[pair] = result.Value
+		timeFilteredPrices[pair.GetOffChainTicker()] = result.Value
```
davidterpay, line 225: Note that the aggregator now stores prices indexed by 
providername -> off-chain-ticker -> price. This greatly reduces the amount of complexity that the aggregator has to deal with with a given market map.

pkg/math/math.go
```diff
@@ -61,3 +63,79 @@ func BigFloatToBigInt(f *big.Float, decimals uint64) *big.Int {
...
+		median = values[middleIndex]
+	}
+
+	return median
+}
+
+// GetScalingFactor returns the scaling factor for the price based on the difference between
+// the token decimals in the erc20 token contracts or similar.
+func GetScalingFactor(
+	first, second int64,
```
davidterpay, line 144: Using int64 instead of uin64 in case of overflow from subtraction

pkg/math/median/median.go
```diff
@@ -2,24 +2,19 @@ package median
...
 )
 
-func ComputeMedianWithContext(_ sdk.Context) types.PriceAggregationFn {
-	return ComputeMedian()
-}
-
 // ComputeMedian inputs the aggregated prices from all providers and computes
 // the median price for each asset.
+//
+// NOTE: This function should only be used for testing purposes.
```
davidterpay, line 13: If we use this in our main aggregation function now, we will effectively be calculating the medians per off-chain ticker and not the cannonical chain representation of the ticker.

providers/apis/dydx/parse.go
```diff
@@ -0,0 +1,216 @@
+package dydx
```
davidterpay, line 1: Moved most of the previous functionality into this file just to keep em separate.

providers/apis/dydx/parse.go
```diff
@@ -0,0 +1,216 @@
...
+	"Huobi":                huobi.Name,
+	"Kucoin":               kucoin.Name,
+	"Okx":                  okx.Name,
+	"Mexc":                 mexc.Name,
+	"CoinbasePro":          coinbase.Name,
+	"TestVolatileExchange": volatile.Name,
+}
+
+// ConvertMarketParamsToMarketMap converts a dYdX market params response to a slinky market map response.
+func (h *APIHandler) ConvertMarketParamsToMarketMap(
```
davidterpay, line 49: We should have a discussion on this.

cmd/slinky-config/main.go
```diff
@@ -458,50 +451,32 @@ func createMarketMap() error {
...
-			// ticker already exists, ensure that the ticker configuration is the same.
-			if t, ok := tickers[tickerStr]; !ok {
-				tickers[tickerStr] = ticker
-			} else if t != ticker {
-				return fmt.Errorf("ticker %s already exists with different configuration for provider %s", tickerStr, name)
+	for _, providerConfig := range ProviderToMarkets {
+		for cp, config := range providerConfig {
+			ticker := mmtypes.Ticker{
+				CurrencyPair:     cp,
+				Decimals:         18,
```
aljo242, line 456: why all 18?

providers/apis/defi/raydium/types.go
```diff
@@ -18,6 +18,25 @@ const (
 	NormalizedTokenAmountExponent = 18
 )
 
+// updateMetaDataCache unmarshals the metadata JSON for each ticker and adds it to the
+// metadata map.
+func (pf *APIPriceFetcher) updateMetaDataCache(ticker types.ProviderTicker) (TickerMetadata, error) {
+	if metadata, ok := pf.metaDataPerTicker[ticker.GetJSON()]; ok {
```
davidterpay, line 25: The primary key is here is the json field, @nivasan1 do you think thats fine or should we just use the offchain ticker?

x/mm2/types/utils.go
```diff
@@ -0,0 +1,30 @@
+package types
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+// ReadMarketConfigFromFile reads a market map configuration from a file at the given path.
+func ReadMarketConfigFromFile(path string) (MarketMap, error) {
```
aljo242: Can we rename this ?

`ReadMarketMapFromFile`

and test?

can just create issue to test at later time

tests/integration/slinky_slashing_setup.go
```diff
@@ -184,32 +183,18 @@ func UpdateNodePrices(node *cosmos.ChainNode, ticker mmtypes.Ticker, price int64
...
+		Markets: map[string]mmtypes.Market{
 			ticker.String(): {
-				Providers: []mmtypes.ProviderConfig{
+				Ticker: ticker,
+				ProviderConfigs: []mmtypes.ProviderConfig{
 					{
 						Name:           static.Name,
-						OffChainTicker: fmt.Sprintf("%d", price),
+						OffChainTicker: ticker.String(),
+						Metadata_JSON:  fmt.Sprintf(`{"price": %f}`, price),
```
aljo242, line 193: `static` provider now works by having a price JSON file as its Metadata JSON

pkg/math/oracle/median.go
```diff
@@ -97,80 +102,76 @@ func (m *MedianAggregator) AggregateData() {
...
 }
 
 // CalculateConvertedPrices calculates the converted prices for a given set of paths and target ticker.
 // The prices utilized are the prices most recently seen by the providers. Each price is within a
 // MaxPriceAge window so is safe to use.
 func (m *MedianAggregator) CalculateConvertedPrices(
 	target mmtypes.Ticker,
-	paths mmtypes.Paths,
-) []*big.Int {
+	providers []mmtypes.ProviderConfig,
```
aljo242: can this function just take in `mmtypes.Market`?

which is just the composition of these two types?

pkg/math/oracle/median.go
```diff
@@ -187,39 +188,22 @@ func (m *MedianAggregator) CalculateConvertedPrices(
...
-
-	// If we have more than one operation, then can only adjust the price using the index.
-	if operations[1].Provider != mmtypes.IndexPrice {
-		return nil, fmt.Errorf("expected index price but got %s", operations[1].Provider)
+	if cfg.NormalizeByPair == nil {
+		return price, nil
 	}
 
-	adjustableByMarketPrice, err := m.GetProviderPrice(operations[1])
+	adjustableByIndexPrice, err := m.GetIndexPrice(*cfg.NormalizeByPair)
```
```suggestion
	normalizeByIndexPrice, err := m.GetIndexPrice(*cfg.NormalizeByPair)
```

oracle/types/oracle.go
```diff
@@ -24,116 +23,121 @@ type (
...
 	// PriceProvider is a type alias for the base price provider. This specifically
 	// implements the provider interface for the price provider along with the
 	// additional base provider methods.
-	PriceProvider = base.Provider[mmtypes.Ticker, *big.Int]
+	PriceProvider = base.Provider[ProviderTicker, *big.Float]
 
 	// PriceProviderI is a type alias for the price provider. This specifically
 	// implements the provider interface for the price provider.
-	PriceProviderI = providertypes.Provider[mmtypes.Ticker, *big.Int]
+	PriceProviderI = providertypes.Provider[ProviderTicker, *big.Float]
```
Eric-Warehime, line 39: These feel very similar and are confusing. Can we consolidate into a single alias/interface for each of these?

oracle/types/oracle.go
```diff
@@ -24,116 +23,121 @@ type (
...
 
 	// PriceAggregationFn is a type alias for the price aggregation function. This function
 	// is used to aggregate the resolved prices from the price providers.
-	PriceAggregationFn = aggregator.AggregateFn[string, TickerPrices]
+	PriceAggregationFn = aggregator.AggregateFn[string, AggregatorPrices]
 
 	// AggregatedProviderPrices is a type alias for the aggregated provider prices. This is
 	// a map of provider names to their respective ticker prices.
-	AggregatedProviderPrices = aggregator.AggregatedProviderData[string, TickerPrices]
+	AggregatedProviderPrices = aggregator.AggregatedProviderData[string, AggregatorPrices]
```
Eric-Warehime, line 114: `map[string]map[string]*big.Float` 

Does it make sense to move all of this stuff into `aggregator/aggregator.go` so we don't have to consult multiple files to figure out this is just a nested map?

oracle/types/oracle.go
```diff
@@ -24,116 +23,121 @@ type (
...
 	// TickerPrices is a type alias for the map of prices. This is a map of tickers i.e.
 	// BTC/USD, ETH/USD, etc. to their respective prices.
-	TickerPrices = map[mmtypes.Ticker]*big.Int
+	TickerPrices = map[ProviderTicker]*big.Float
+
+	// AggregatorPrices is a type alias for a map of off-chain ticker to the price. It is
+	// critical to note that the aggregator tracks prices indexed by off-chain tickers. When
+	// a final price is calculated (after aggregation), it is indexed by the canonical on-chain
+	// ticker.
+	AggregatorPrices = map[string]*big.Float
```
g3ndo9kar9, line 101: Can we call this ProviderPrices or smth so it indicates where this is used

oracle/types/provider.go
```diff
@@ -0,0 +1,103 @@
...
+		GetOffChainTicker() string
+		// GetJSON returns additional JSON data for the ticker.
+		GetJSON() string
+	}
+
+	// DefaultProviderTicker is a basic implementation of the ProviderTicker interface.
+	// Provider's that utilize this implementation should be able to easily configure
+	// custom json data for their tickers.
+	DefaultProviderTicker struct {
+		Name           string
```
g3ndo9kar9: What does Name correspond to?

g3ndo9kar9: The name of the Provider?

oracle/types/provider.go
```diff
@@ -0,0 +1,103 @@
...
+	"sync"
+)
+
+type (
+	// ProviderTicker is the interface for the ticker that provider's utilize/return.
+	ProviderTicker interface {
+		fmt.Stringer
+
+		// GetProvider returns the provider for the ticker.
+		GetProvider() string
```
Eric-Warehime: Why does the ProviderTicker need to know which provider it's associated with? I can't see anywhere that the `GetProvider` method is called, and when `ProviderTicker` is used it's generally in the context of a `ProviderTickers` which is just an association between a single Provider and its offchain Tickers.

pkg/math/median/median.go
g3ndo9kar9: Let's move this to a test-utils

aljo242 (approved):

Eric-Warehime (approved):

PR #325, opened 2024-04-05
refactor: API Price Providers with Provider Ticker Interface (PR2)

davidterpay:


oracle/types/provider.go
```diff
@@ -13,24 +13,24 @@ type (
...
 	// custom json data for their tickers.
 	DefaultProviderTicker struct {
-		provider string
-		offChain string
-		decimals uint64
-		json     string
+		Name           string
+		OffChainTicker string
+		Decimals       uint64
+		JSON           string
```
davidterpay: Doing this so that all of our provider market configs can be backwards compatible.

oracle/types/market.go
```diff
@@ -64,3 +64,15 @@ func (tpt *TickersToProviderTickers) ToProviderTickers() []ProviderTicker {
 
 	return providerTickers
 }
+
+// MustGetProviderTicker returns the provider ticker for the given x/marketmap ticker.
+// This function is mostly used for testing.
+func (tpt *TickersToProviderTickers) MustGetProviderTicker(ticker mmtypes.Ticker) ProviderTicker {
```
aljo242: seems like this should just not be a pointer receiver function

pkg/math/math.go
```diff
@@ -51,3 +52,33 @@ func BigFloatToBigInt(f *big.Float, decimals uint64) *big.Int {
 
 	return result
 }
+
+// Float64StringToBigFloat converts a float64 string to a big.Float.
+func Float64StringToBigFloat(s string, decimals uint64) (*big.Float, error) {
```
aljo242: Are we going to remove the `BigInt` functions?  

Or can we express them in terms of the `BigFloat` functions if we keep them

pkg/math/math.go
```diff
@@ -51,3 +52,33 @@ func BigFloatToBigInt(f *big.Float, decimals uint64) *big.Int {
...
+	_, ok := bigFloat.SetString(s)
+	if !ok {
+		return nil, fmt.Errorf("failed to set big.Float from string: %s", s)
+	}
+
+	return ScaleBigFloat(bigFloat, decimals), nil
+}
+
+// Float64ToBigFloat converts a float64 to a big.Float.
+func Float64ToBigFloat(val float64, decimals uint64) *big.Float {
```
aljo242: test

pkg/math/math.go
```diff
@@ -51,3 +52,33 @@ func BigFloatToBigInt(f *big.Float, decimals uint64) *big.Int {
...
+	bigVal := new(big.Float)
+	bigVal.SetFloat64(val)
+
+	return ScaleBigFloat(bigVal, decimals)
+}
+
+// ScaleBigFloat scales a big.Float by the given decimals.
+func ScaleBigFloat(f *big.Float, decimals uint64) *big.Float {
```
aljo242, line 67: test

providers/apis/kraken/api_handler.go
```diff
@@ -20,25 +19,16 @@ var _ types.PriceAPIDataHandler = (*APIHandler)(nil)
 // for more information about the Kraken API, refer to the following link:
 // https://docs.kraken.com/rest/
 type APIHandler struct {
-	// market is the config for the Kraken API.
-	market types.ProviderMarketMap
 	// api is the config for the Kraken API.
 	api config.APIConfig
+	// cache maintains the latest set of tickers seen by the handler.
+	cache types.ProviderTickers
```
aljo242, line 25: One thing im thinking while reading this PR:

Does this cache need any eviction?

oracle/types/market.go
davidterpay: Is there a reason why?

pkg/math/math.go
davidterpay: I think we keep the BigInt functions till we have everything converted to BigFloat.

providers/apis/kraken/api_handler.go
davidterpay: The cache is technically reset everytime the provider is so this should be safe, in terms of how the cache is actually going to operate if there are multiple of the same off-chain tickers is that the latest one will be stored which is fine. We do now have a uniqueness constraint on the off-chain-ticker tho.

oracle/types/market.go
```diff
@@ -53,14 +53,34 @@ func ProviderTickersFromMarketMap(
 type TickersToProviderTickers map[mmtypes.Ticker]DefaultProviderTicker
 
 // ToProviderTickers converts the map to a list of provider tickers.
-func (tpt *TickersToProviderTickers) ToProviderTickers() []ProviderTicker {
-	var providerTickers = make([]ProviderTicker, len(*tpt))
+func (tpt TickersToProviderTickers) ToProviderTickers() []ProviderTicker {
+	var providerTickers = make([]ProviderTicker, len(tpt))
```
Eric-Warehime, line 56: `make format`

oracle/types/market.go
```diff
@@ -53,14 +53,34 @@ func ProviderTickersFromMarketMap(
...
 
 	i := 0
-	for _, ticker := range *tpt {
+	for _, ticker := range tpt {
+		// If the ticker does not have a set number of decimals, set it to the default.
+		if ticker.Decimals == 0 {
+			ticker.Decimals = DefaultTickerDecimals
+		}
```
Eric-Warehime: Isn't this prevented during initialization in ValidateBasic?

oracle/types/market.go
```diff
@@ -53,14 +53,34 @@ func ProviderTickersFromMarketMap(
...
 
 	return providerTickers
 }
+
+// MustGetProviderTicker returns the provider ticker for the given x/marketmap ticker.
+// This function is mostly used for testing.
+func (tpt TickersToProviderTickers) MustGetProviderTicker(ticker mmtypes.Ticker) ProviderTicker {
+	providerTicker, ok := (tpt)[ticker]
+	if !ok {
+		panic(fmt.Sprintf("ticker %s not found", ticker))
```
Eric-Warehime, line 72: Kind of seems like a footgun having this panic.

aljo242 (approved):

oracle/types/market.go
davidterpay: I only use this in testing so it's probably fine?

oracle/constants/tickers.go
```diff
@@ -1,7 +1,7 @@
 package constants
 
 import (
-	mmtypes "github.com/skip-mev/slinky/x/marketmap/types"
+	mmtypes "github.com/skip-mev/slinky/x/mm2/types"
```
g3ndo9kar9, line 4: kek

oracle/types/market.go
g3ndo9kar9: Agreed, I think we should fail a market / provider if a ticker is not present? And we should return an error here (or we should just error on addition of invalid tickers)

oracle/types/market.go
```diff
@@ -53,14 +53,34 @@ func ProviderTickersFromMarketMap(
 type TickersToProviderTickers map[mmtypes.Ticker]DefaultProviderTicker
```
g3ndo9kar9, line 52: Why use a `Ticker` as the index to a data-structure used in providers? The entire point here is to decouple providers / market-map types? If we need this structure, why can't the index be the offChainTicker?

oracle/types/market.go
g3ndo9kar9: `Must` prefices are p standard in cosmos, and they always panic instead of error. If we only use in tests, can we just move this to testing, and have it wrap the existing getter?

oracle/types/oracle.go
```diff
@@ -23,20 +23,20 @@ type (
...
 	// PriceProviderFactory is a type alias for the price provider factory. This
 	// specifically only returns price providers that implement the provider interface.
-	PriceProviderFactoryI = factory.ProviderFactory[ProviderTicker, *big.Int]
+	PriceProviderFactoryI = factory.ProviderFactory[ProviderTicker, *big.Float]
 
 	// PriceProvider is a type alias for the base price provider. This specifically
 	// implements the provider interface for the price provider along with the
 	// additional base provider methods.
-	PriceProvider = base.Provider[ProviderTicker, *big.Int]
+	PriceProvider = base.Provider[ProviderTicker, *big.Float]
```
g3ndo9kar9, line 35: kek

oracle/types/provider.go
```diff
@@ -13,28 +14,30 @@ type (
...
-		// Decimals returns the number of decimals for the ticker.
-		Decimals() uint64
-		// JSON returns additional JSON data for the ticker.
-		JSON() string
+		// GetProvider returns the provider for the ticker.
+		GetProvider() string
+		// GetOffChainTicker returns the off-chain representation for the ticker.
+		GetOffChainTicker() string
+		// GetDecimals returns the number of decimals for the ticker.
+		GetDecimals() uint64
```
g3ndo9kar9: Why does a provider need Decimals?

pkg/math/math.go
g3ndo9kar9: Why do we need this? Like the entire purpose of using `big.Floats` is that they don't need to be scaled for arithmetic?

pkg/math/math.go
g3ndo9kar9: Why does this take  decimals? Any time we're scaling a value by an exponent, we have to be aware of the scaling factor in subsequent arithmetic operators, IMO moving to `big.Float` lets us avoid this

providers/apis/binance/api_handler.go
```diff
@@ -106,12 +93,12 @@ func (h *APIHandler) ParseResponse(
 
 	for _, data := range result {
 		// Filter out the responses that are not expected.
-		ticker, ok := h.market.OffChainMap[data.Symbol]
+		ticker, ok := h.cache.FromOffChainTicker(data.Symbol)
```
g3ndo9kar9, line 96: We're introducing a cache consistency problem here, no? If the cached tickers vary from the given tickers we can have an inconsistency. If we're caching all ticker state to avoid O(n) lookups, I'd either update the cache in `ParseResponse` or pass a set, instead of an array of `ProviderTickers` to `ParseResponse`

g3ndo9kar9 (commented):
comments

oracle/types/market.go
davidterpay: We still need a mapping of on-chain to off-chain rep so that our providers can have local configs that we can use for testing and such. I do think that we could probably simplify it to use just currencypair which omits much of the other info that Ticker includes.

providers/apis/binance/api_handler.go
davidterpay: CreateURL will always be called before ParseResponse but im fine moving the caching from CreateURL to here. seems like the safest option.

PR #332, opened 2024-04-09
feat(providers/apis): Add optional authentication per endpoint [BLO-1069]

g3ndo9kar9:
## In This PR
- I introduce a `RoundTripper` implementation which will introduce a configured set of headers to outbound requests
     - The expected header fields are for authentication per rpc-provider
     - The `RoundTripper` interface is general enough for us to use across all of our providers (as long as they depend on the `http.Client`
- I update the raydium provider to configure the `*http.Client` used based on this transport implementation


oracle/config/api.go
```diff
@@ -50,6 +50,43 @@ type APIConfig struct {
 type Endpoint struct {
 	// URL is the URL that is used to fetch data from the API.
 	URL string `json:"url"`
+
+	// Authentication  is used to meta-data related to authentication for the API's
```
aljo242: This godoc doesn't make sense

oracle/config/api_test.go
```diff
@@ -148,6 +148,49 @@ func TestAPIConfig(t *testing.T) {
 			},
 			expectedErr: false,
 		},
+		{
+			name: "bad config with invalid endpoint",
```
aljo242: "invalid endpoint authentication"

g3ndo9kar9: wym

aljo242: the name for the test could be more descriptive

aljo242 (approved):

oracle/config/api.go
```diff
@@ -50,6 +50,50 @@ type APIConfig struct {
...
+
+	return e.Authentication.ValidateBasic()
+}
+
+// Authentication holds all data necessary for an API provider to authenticate with an
+// endpoint.
+type Authentication struct {
+	// Enabled is a flag that indicates whether the provider should authenticate with
+	// the endpoint.
+	Enabled bool `json:"enabled"`
```
Eric-Warehime: Do we need this or can we just use the auth info if it's present and assume it's un-auth'd if there is no auth info.

Eric-Warehime (approved):
LGTM, I just don't like having enabled flags when we can just either have the data specified or not.

oracle/config/api.go
g3ndo9kar9: I can remove

PR #337, opened 2024-04-10
chore: Cleaning up deadcode

davidterpay:


pkg/math/oracle/aggregator.go
```diff
@@ -1,79 +1,209 @@
 package oracle
 
 import (
-	"maps"
+	"fmt"
```
davidterpay, line 4: No new code, just renaming and moving to a different file name since it was previously in median.go

aljo242 (approved):

PR #343, opened 2024-04-11
fix: Default accept nil VE

davidterpay:


aljo242:
@davidterpay do we need this PR?

Eric-Warehime (approved):

aljo242 (approved):

abci/ve/vote_extension.go
```diff
@@ -232,6 +232,17 @@ func (h *VoteExtensionHandler) VerifyVoteExtensionHandler() sdk.VerifyVoteExtens
 			h.logger.Error("VerifyVoteExtensionHandler received a nil request")
 			return nil, err
 		}
+
+		// By default, we accept empty vote extensions.
+		if len(req.VoteExtension) == 0 {
```
g3ndo9kar9, line 237: We are we only doing this here, and not replicating in Prepare / Process, etc?

g3ndo9kar9:
If we want to explicitly accept NilVoteExtensions, I'd move this logic to our codecs? That way we aren't duplicating everywhere we have to verify a vote-extension

davidterpay:
> If we want to explicitly accept NilVoteExtensions, I'd move this logic to our codecs? That way we aren't duplicating everywhere we have to verify a vote-extension

@nivasan1 isn't the top level verifyVE the best place to do this? wouldn't we have to replicate this per codec otherwise

PR #344, opened 2024-04-11
feat(providers/api/marketmap): Add raydium to dydx market-map provider [BLO-1084]

g3ndo9kar9:


providers/apis/dydx/parse.go
```diff
@@ -180,35 +185,118 @@ func (h *APIHandler) ConvertExchangeConfigJSON(
...
+			return "", fmt.Errorf("failed to marshal %s provider metadata for ticker %s: %w", raydium.Name, ticker, err)
+		}
+
+		return string(bz), nil
+	}
+
+	return "", nil
+}
+
+// raydiumTickerMetadata extracts a raydiumTickerMetadata from a dydx ticker.
```
davidterpay: can we move this to the raydium provider and have it be a helper function.

davidterpay: Can we also document what the expected ticker string looks like?

providers/apis/dydx/parse.go
```diff
@@ -180,35 +185,118 @@ func (h *APIHandler) ConvertExchangeConfigJSON(
...
+func raydiumTickerMetadata(ticker string) (raydium.TickerMetadata, error) {
+	// split fields by separator and expect there to be at least 6 values
+	fields := strings.Split(ticker, "/")
```
davidterpay: dYdX uses "-" as a delimeter btw.

davidterpay: We should assume that

providers/apis/dydx/parse.go
```diff
@@ -180,35 +185,118 @@ func (h *APIHandler) ConvertExchangeConfigJSON(
...
 			Invert:          cfg.Invert,
 			NormalizeByPair: normalizeByPair,
+			Metadata_JSON:   metaData,
 		})
 	}
 
 	return providers, nil
 }
 
+// ExtractMetadataFromTicker extracts json-metadata from a ticker, based on the exchange. Specifically,
```
davidterpay: This is likely not going to be raydium specific, could we move these docs to raydium

aljo242 (approved):
Yay

Eric-Warehime (approved):

PR #353, opened 2024-04-18
chore: reduce log level in providers

aljo242:
- move `Info` -> `Debug` wherever makes sense
- move `Error` -> `Debug` whenever we return an error with the same message


providers/apis/marketmap/api_handler.go
```diff
@@ -22,7 +22,7 @@ type APIHandler struct {
 	api config.APIConfig
 }
 
-// NewAPIHandler returns a new MarketMap MarketMapAPIDataHandler.
+// NewAPIHandler returns a new MarketMap MarketMapAPIDatzaHandler.
```
Eric-Warehime: this is not correct

providers/apis/defi/raydium/multi_client.go
```diff
@@ -90,7 +90,7 @@ func (c *MultiJSONRPCClient) GetMultipleAccountsWithOpts(
 			defer wg.Done()
 			resp, err := client.GetMultipleAccountsWithOpts(ctx, accounts, opts)
 			if err != nil {
-				c.logger.Error("failed to fetch accounts", zap.Error(err))
```
Eric-Warehime: It feels like we should have some error logging here if we're unable to produce any price for raydium. I think it's ok for debug logs on individual client errors, but if the whole raydium provider is not producing any prices we probably want some logs. Would we still get error logs somewhere else if this is a debug?

providers/apis/defi/uniswapv3/fetcher.go
```diff
@@ -110,7 +110,7 @@ func (u *PriceFetcher) Fetch(
 	for i, ticker := range tickers {
 		pool, err := u.GetPool(ticker)
 		if err != nil {
-			u.logger.Error(
```
Eric-Warehime, line 113: Similar comment as for raydium--it seems like we should have some error log (only 1 though) when we fail outright. Maybe we still will though?

providers/apis/defi/raydium/multi_client.go
aljo242: yeah this makes sense.  I will make this Error for now.  I think we should revisit this and probably accrue errors to pass up the chain like we do in other providers here

providers/apis/defi/uniswapv3/fetcher.go
aljo242: here - we will have an error log, since we're returning the error.  Eventually it will be logged

Eric-Warehime (approved):

PR #354, opened 2024-04-19
feat: Add dynamic provider naming to uniswapv3

Eric-Warehime:
Adds dynamic provider naming for uniswapv3.

The provider name will now be required to be `BaseName` `Separator` `Chain` where `Separator` is set to `-`, `BaseName` is `uniswapv3`, and `Chain` is the lowercase chain name. 

For our only supported configuration currently that gives us:

`uniswapv3-ethereum` for uniswapv3 on ethereum mainnet.

Also adds in the plubming for uniswapv3-ethereum in dydx. We will expect dydx's `ExchangeMarketConfigJson` to contain a `Ticker` field with `PoolContractAddress` `-` `BaseDecimals` `-` `QuoteDecimals`

Examples are included in the tests.


aljo242 (approved):
lgtm

PR #358, opened 2024-04-19
feat: Ethmulticlient with auth

Eric-Warehime:
Adds an ethereum client implementation which calls multiple underlying clients.

Each one can have its own authentication specified in the oracle.json config.  
The multi client injects a `eth_blockNumber` call into the set of batched calls in order to determine which response is the freshest. If all responses fail, we bubble up an error, otherwise we return the response from the first node we see at the highest observed height.


providers/apis/defi/ethmulticlient/multi_client.go
```diff
@@ -0,0 +1,111 @@
...
+func NewMultiRPCClientFromEndpoints(ctx context.Context, logger *zap.Logger, endpoints []config.Endpoint) (*MultiRPCClient, error) {
+	if logger == nil {
+		return nil, fmt.Errorf("logger cannot be nil")
+	}
+	clients := make([]EVMClient, len(endpoints))
+	for i, endpoint := range endpoints {
+		var err error
+		clients[i], err = NewGoEthereumClientImplFromEndpoint(ctx, endpoint)
+		if err != nil {
+			logger.Error(
```
aljo242, line 41: do we need this log?

aljo242 (approved):

providers/apis/defi/ethmulticlient/multi_client.go
Eric-Warehime: This should only be called on provider initialization so it seems ok to log this when we fail to initialize the provider?

PR #369, opened 2024-04-24
chore: Updating metrics

davidterpay:


Makefile
```diff
@@ -52,11 +52,11 @@ update-local-configs: build
...
 	@./build/slinky-config --chain dydx --oracle-config-path ${DYDX_ORACLE_CONFIG_FILE} --node-http-url=localhost:1317 --raydium-enabled=true --solana-node-endpoint ${SOLANA_NODE_ENDPOINT}
 
-start-oracle:
+start-all:
 	@echo "Starting oracle side-car, blockchain, and prometheus dashboard..."
 	@$(DOCKER_COMPOSE) -f docker-compose.yml up -d
 
-stop-oracle:
+stop-all:
 	@echo "Stopping network..."
```
Eric-Warehime: Why do we need to change this?

Eric-Warehime (approved):
Metrics changes lgtm

Makefile
davidterpay: we have a cmd `start-sidecar` which kinda clashes with `start-oracle`. the latter also spins up a local node

service/metrics/metrics.go
```diff
@@ -169,14 +170,14 @@ func (m *metricsImpl) ObserveMessageSize(messageType MessageType, size int) {
 func (m *metricsImpl) ObservePriceForTicker(ticker slinkytypes.CurrencyPair, price float64) {
 	m.prices.With(prometheus.Labels{
 		ChainIDLabel: m.chainID,
-		TickerLabel:  ticker.String(),
+		TickerLabel:  strings.ToLower(ticker.String()),
```
aljo242, line 173: I feel like we should be doing the opposite here.

On chain the tickers are BTC/USD.  and also internally in the sidecar we format them like that.

I feel like the metrics should be using the canonical way we represent tickers

davidterpay: side car lower cases everything

davidterpay: along with metrics

davidterpay: we're going to have to make many more changes to get everything to upper-case, i prefer just sticking with this unless the vals are very opinionated

aljo242: I would still prefer to upper case everything because that unifies everything.

Say I'm looking at metrics and see the price for `btc/usd`, then i run an on-chain query for it using "btc/usd".  It will appear as thought the price isnt there.

aljo242: If this seems unnecessary or too much work im fine with not doing this tho.  its not too big of a deal or vals don't care

davidterpay: yea i'd prefer to discuss this internally before we go with it. down for either, the only hesitation is that it'll require more changes.

davidterpay: probably best for a separate task.

aljo242 (approved):

service/metrics/metrics.go
```diff
@@ -76,7 +77,7 @@ func NewMetrics(chainID string) Metrics {
 			Help:      "The time it took for an ABCI method to execute slinky specific logic (in seconds)",
 			Buckets:   []float64{.0001, .0004, .002, .009, .02, .1, .65, 2, 6, 25},
 		}, []string{ABCIMethodLabel, ChainIDLabel}),
-		abciRequests: prometheus.NewCounterVec(prometheus.CounterOpts{
+		abciRequests: prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
g3ndo9kar9, line 80: Have we checked that success / failure are starting at 0?

PR #371, opened 2024-04-24
feat: Add core configs

Eric-Warehime:
Adds a config directory for our core chains. The `oracle.json` is an exact copy from the `local` config for now.

The `market.json` is a copy of the `dydx` `market.json` with the following changes:
* `min_provider_count` for `USDT/USD` is reduced to 1
This currency pair is required for normalizing nearly every single price we fetch. It therefore controls whether or not we post prices for every market, and it feels correct to favor posting participation of validators over price liveness for this-especially since the price in theory won't move very much.

* NTRN/USD market added
Tested this out locally and all of the markets work. I wasn't able to test binance since it's not available in my location, and binance.us doesn't have ntrn, but the regular binance site has a market for it with plenty of volume.

I haven't added any of the `slinky-config` generation stuff since this market map should be relatively static, and dynamic market maps will soon be generated via the MMU.


aljo242:
@Eric-Warehime can we set this up as the default market map to run when we run our `docker-compose` local env ?

docker-compose.yml
```diff
@@ -8,8 +8,8 @@ services:
       context: .
       dockerfile: contrib/images/slinky.sidecar.prod.Dockerfile
     volumes:
-      - ./config/dydx/oracle.json:/oracle/oracle.json
-      - ./config/dydx/market.json:/oracle/market.json
+      - ./config/core/oracle.json:/oracle/oracle.json
+      - ./config/core/market.json:/oracle/market.json
```
g3ndo9kar9: Is it possible to template this? I frequently have to edit this myself depending on what I'm testing

g3ndo9kar9 (approved):
lgtm

aljo242 (approved):

PR #374, opened 2024-04-25
feat: Log Reduction + Level Logging

davidterpay:


oracle/config/oracle.go
```diff
@@ -25,6 +25,8 @@ type OracleConfig struct {
 
 	// Production specifies whether the oracle is running in production mode. This is used to
 	// determine whether the oracle should be run in debug mode or not.
+	//
+	// TODO:  Deprecate this field in favor of a more robust configuration system.
```
davidterpay: Im assuming we dont want to deprecate this till a major release?

aljo242: I think we can deprecate now and then remove at a later time.

can add a godoc like [this](https://zchee.github.io/golang-wiki/Deprecated/)

aljo242 (approved):
lgtm - i think we should just deprecate the `Production` field now

cmd/slinky/main.go
```diff
@@ -77,6 +80,20 @@ func init() {
...
+	rootCmd.Flags().StringVarP(
+		&logLevel,
+		"log-level",
+		"",
+		"info",
+		"Log level (debug, info, warn, error, dpanic, panic, fatal).",
+	)
+	rootCmd.Flags().StringVarP(
+		&writeLogsTo,
+		"write-logs-to",
```
Eric-Warehime: Can we make this `log-file`, that way we have
`--log-level` and `--log-file`.

pkg/log/zap.go
```diff
@@ -0,0 +1,80 @@
...
+	if config.WriteTo != "" {
+		outputPaths = append(outputPaths, config.WriteTo)
+	}
+
+	// initialFields are the fields that are added to every log message.
+	initialFields := map[string]interface{}{
+		"pid": os.Getpid(),
+	}
+	if config.WriteTo != "" {
+		initialFields["writing_logs_to"] = config.WriteTo
```
Eric-Warehime: This means every log will have a property like the following:
`"writing_logs_to": "<the file you're currently looking at>"` right?

Eric-Warehime (approved):
LGTM--small comments

pkg/log/zap.go
davidterpay: now that i think about it we probably should have it be a part of every log.

davidterpay: removed it so it is not part of every log

cmd/slinky/main.go
```diff
@@ -77,6 +80,20 @@ func init() {
...
 		"Port for the pprof server to listen on.",
 	)
+	rootCmd.Flags().StringVarP(
+		&logLevel,
+		"log-level",
+		"",
+		"info",
+		"Log level (debug, info, warn, error, dpanic, panic, fatal).",
+	)
+	rootCmd.Flags().StringVarP(
```
g3ndo9kar9, line 90: Why do we need this? @Eric-Warehime is this standard practice? My intuition is that all infra deployments assume logs are piped to stdout, and will redirect wherever they want from there (i.e in solve)

g3ndo9kar9: see [here](https://github.com/skip-mev/solve/blob/main/logging/logging.go#L21)

cmd/slinky/main.go
```diff
@@ -109,18 +126,12 @@ func runOracle() error {
...
-		if err != nil {
-			return fmt.Errorf("failed to create logger: %s", err.Error())
-		}
-	}
+	// Build logger.
+	logger := log.NewLogger(log.Config{
+		LogLevel: logLevel,
+		WriteTo:  writeLogsTo,
+	})
+	defer logger.Sync()
```
g3ndo9kar9, line 134: What does this do?

pkg/log/zap.go
```diff
@@ -0,0 +1,77 @@
...
+	case "error":
+		logLevel = zapcore.ErrorLevel
+	case "dpanic":
+		logLevel = zapcore.DPanicLevel
+	case "panic":
+		logLevel = zapcore.PanicLevel
+	case "fatal":
+		logLevel = zapcore.FatalLevel
+	default:
+		logLevel = zapcore.InfoLevel
```
g3ndo9kar9, line 60: Can't we get all of this directly from [zap](https://arc.net/l/quote/yccvklvf)

pkg/log/zap.go
```diff
@@ -0,0 +1,77 @@
...
+	// outputPaths is the list of paths to write logs to. By default, logs are
+	// written to stderr.
+	outputPaths := []string{"stderr"}
+	if config.WriteTo != "" {
+		outputPaths = append(outputPaths, config.WriteTo)
+	}
+
+	// initialFields are the fields that are added to every log message.
+	initialFields := map[string]interface{}{
+		"pid": os.Getpid(),
```
g3ndo9kar9, line 36: Why do we need this? Aren't all logs logged from the same process anyway?

pkg/log/zap.go
```diff
@@ -0,0 +1,77 @@
...
+		logLevel = zapcore.PanicLevel
+	case "fatal":
+		logLevel = zapcore.FatalLevel
+	default:
+		logLevel = zapcore.InfoLevel
+	}
+
+	zapConfig := zap.Config{
+		Level:             zap.NewAtomicLevelAt(logLevel),
+		Development:       dev,
```
g3ndo9kar9, line 65: What does this field mean?

g3ndo9kar9 (commented):
comments

cmd/slinky/main.go
davidterpay: more of a utility thing than anything else

cmd/slinky/main.go
davidterpay: blocks on process closing till all logs are written

pkg/log/zap.go
davidterpay: utility field more than anything else

davidterpay: otherwise you have to lsof

pkg/log/zap.go
davidterpay: prints more in depth stack traces in the case where it is dev

PR #390, opened 2024-05-03
feat: Separate ws subscribe requests

Eric-Warehime:
This separates all of the subscribe messages for our websocket providers into different messages. 

## Why
Almost all of the websocket providers batch every ticker they need to fetch into a single subscribe request. The result of this is that if a single market in that request fails, every one of the markets a provider is responsible for will not be fetched.

## What
I've moved each market to a separate subscription request. This means that after a provider is spun up and a connection is created, the provider will attempt to make X Subscribe requests on that connection, where X is the number of markets it is responsible for fetching.

Our ws providers do not open multiple connections (afaict in this code path) so this change does not affect the number of open tcp connections and so should have zero impact on performance outside of slower startup time for fetching prices and slightly higher network consumption because of duplicated data sent during multiple subscribe requests that were previously being batched.


aljo242 (approved):

Eric-Warehime:
Dashboard looks good w/ these changes.  
https://skipprotocol.grafana.net/dashboard/snapshot/2N7xOugZDVH7SnZxlc0XO8A5LJih9v9Q

PR #391, opened 2024-05-03
feat: DYDX research-JSON market-map provider [BLO-1173]

g3ndo9kar9:
## In This PR
- I introduce a new config for dydx `dydx_research`, which contains all research markets and will be updated dynamically as new markets are added to the dydx [research-json](https://raw.githubusercontent.com/dydxprotocol/v4-web/main/public/configs/otherMarketData.json)
- I introduce a `research_api_handler` which supplements the current dydx market-map api-handler to add all markets from the research-json in each update to the dydx mainnet market-map. Specifically,
     - I implement a new `Fetcher` implementation which will wrap the dydx `api_handler` and `research_api_handler` to concurrently fetch for both market-maps, convert, and combine the markets as follows
         - If a market exists in the research-markets that's not included in mainnet, add the new market to the mainnet market-map
 - I add a `dydxResearchJSONMarketMap` flag to the `slinky-config` generation script (+ Makefile target) which will signal to generate the `dydx_research` `oracle.json`
       - The default configs point to to the dydx research-json on github + lavendar 5's dydx mainnet node for the mainnet market-map
## Let's Discuss
- I'll be adding a permanent staging environment for this sidecar tagged to the latest commit-hash in this PR
[dashboard](https://skipprotocol.grafana.net/d/cdca49ue665mocd/slinky-sidecar-dashboard?orgId=1&var-DS_PROMETHEUS=default&var-job=slinky-preview-pr-391-sidecar-0&var-id=All&var-provider=All&var-market_map_provider=dydx_api&from=now-5m&to=now)


cmd/slinky-config/main.go
```diff
@@ -275,6 +278,15 @@ func init() {
 		"",
 		"Chain that we expect the oracle to be running on {dydx, \"\"}. This should only be specified if required by the chain.",
 	)
+
+	rootCmd.Flags().BoolVarP(
+		&dydxResearchJSONMarketMap,
+		"dydx-research-json-market-map",
+		"",
+		false,
+		"Whether to use the dydx-research json to configure markets alongside the chain.",
```
aljo242:
```suggestion
		"Use the dydx-research json to configure markets alongside the chain.",
```

pkg/arrays/arrays.go
```diff
@@ -0,0 +1,12 @@
+package arrays
+
+// CheckEntryInArray checks if an entry is in an array, and returns a boolean value
```
aljo242:
```suggestion
// CheckEntryInArray checks if an entry is in an array, and returns true
```

pkg/arrays/arrays_test.go
```diff
@@ -0,0 +1,45 @@
+package arrays_test
+
+import (
+	"testing"
+
+	"github.com/skip-mev/slinky/pkg/arrays"
+)
+
+func TestCheckEntryInArray(t *testing.T) {
+	t.Parallel()
```
aljo242, line 10: not needed

aljo242 (approved):

pkg/arrays/arrays_test.go
g3ndo9kar9: Our linter pointed this out

Eric-Warehime (approved):
lgtm, not sure if we really need a multi mm fetcher--feels like just running the research json or the mainnet separately would be fine also.

providers/apis/dydx/utils.go
```diff
@@ -52,6 +55,23 @@ var DefaultAPIConfig = config.APIConfig{
...
 
+// DefaultResearchAPIConfig returns the default configuration for the dYdX market map API.
+var DefaultResearchAPIConfig = config.APIConfig{
+	Name:             ResearchAPIHandlerName,
+	Atomic:           true,
+	Enabled:          true,
+	Timeout:          20 * time.Second, // Set a high timeout to account for slow API responses in the case where many markets are queried.
+	Interval:         10 * time.Second,
+	ReconnectTimeout: 2000 * time.Millisecond,
+	URL:              "https://dydx-api.lavenderfive.com:443",
```
davidterpay, line 66: We want this to be on our main?

davidterpay: I'm assuming we should shield all val specific fields from being in the repo.

config/dydx_research/oracle.json
```diff
@@ -0,0 +1,541 @@
...
+    {
+      "name": "dydx_research_json_api",
+      "api": {
+        "enabled": true,
+        "timeout": 20000000000,
+        "interval": 10000000000,
+        "reconnectTimeout": 2000000000,
+        "maxQueries": 1,
+        "atomic": true,
+        "url": "https://dydx-api.lavenderfive.com:443",
```
davidterpay, line 462: same applies here, we shouldnt put val url's in our configs.

davidterpay:
I would remove the val url, otherwise lgtm

davidterpay (approved):

PR #398, opened 2024-05-07
chore: Move provider count tracking

davidterpay:
moving where we collect the number of providers used for aggregation to be a bit earlier in price calculation since we otherwise won't collect the metric.


aljo242:
Can you give more context on this change? what is this doing?

g3ndo9kar9 (approved):
kek

PR #399, opened 2024-05-07
feat: use the enabled field from `MarketMap`

aljo242:
Closes BLO-1089


Makefile
```diff
@@ -168,8 +168,11 @@ build-configs:
 	@jq '.consensus["params"]["abci"]["vote_extensions_enable_height"] = "2"' $(GENESIS) > $(GENESIS_TMP) && mv $(GENESIS_TMP) $(GENESIS)
 	@jq '.app_state["oracle"]["currency_pair_genesis"] += [{"currency_pair": {"Base": "BTC", "Quote": "USD"},"currency_pair_price": null,"nonce": "0"}]' $(GENESIS) > $(GENESIS_TMP) && mv $(GENESIS_TMP) $(GENESIS)
 	@jq '.app_state["oracle"]["next_id"] = "2"' $(GENESIS) > $(GENESIS_TMP) && mv $(GENESIS_TMP) $(GENESIS)
+	@dasel put -r toml 'rpc.laddr' -f $(CONFIG_TOML) -t string -v "tcp://0.0.0.0:26657"
 	@dasel put -r toml 'telemetry.enabled' -f $(APP_TOML) -t bool -v true
-	@dasel put -r toml 'instrumentation.enabled' -f $(CONFIG_TOML) -t bool -v true
+	@dasel put -r toml 'api.enable' -f $(APP_TOML) -t bool -v true
+	@dasel put -r toml 'grpc.address' -f $(APP_TOML) -t string -v "0.0.0.0:9090"
+	@dasel put -r toml 'api.address' -f $(APP_TOML) -t string -v "tcp://0.0.0.0:1317"
```
aljo242, line 175: this allows us to curl query the node

config/dydx/market.json
```diff
@@ -7,7 +7,8 @@
           "Quote": "USD"
         },
         "decimals": 8,
-        "min_provider_count": 3
+        "min_provider_count": 3,
+        "enabled": true
```
aljo242, line 11: enable everything

oracle/types/market.go
```diff
@@ -28,6 +24,10 @@ func ProviderTickersFromMarketMap(
 	// Iterate through every single market and its provider configurations to find the
 	// provider configurations that match the provider name.
 	for _, market := range marketMap.Markets {
+		if !market.Ticker.Enabled {
```
aljo242, line 27: this is actually where we skip the disabled providers

oracle/types/market_test.go
```diff
@@ -26,23 +26,32 @@ func TestProviderTickersFromMarketMap(t *testing.T) {
 			err:      false,
 		},
 		{
-			name:     "invalid market map",
+			name:     "single disabled market",
```
Eric-Warehime, line 29: I think the thing which is missing here for me is:
* what do we do if there is a transitive dependency on a disabled market?
For example, do we allow `USDT/USDC` to be disabled if that's our normalize market for other things?

providers/apis/dydx/parse.go
```diff
@@ -115,6 +115,7 @@ func (h *APIHandler) CreateTickerFromMarket(market dydxtypes.MarketParam) (mmtyp
 		CurrencyPair:     cp,
 		Decimals:         uint64(market.Exponent * -1),
 		MinProviderCount: uint64(market.MinExchanges),
+		Enabled:          true,
```
aljo242, line 118: We always enable everything parsed from dydx

oracle/types/market_test.go
davidterpay: I was thinking the same thing, if a given feed should be accessible but the overall market shouldn't be calculated. Is this a case we should be worried about?

Makefile
davidterpay: what is this doing?

davidterpay:
Adding a preview for this to see how it runs

Makefile
Eric-Warehime: I think the sdk defaults the node's API endpoint to off unless you set `api.enable` and `api.address` I guess is the port it's serving on. Not sure what the default value for that is--thought it was `localhost:1317` which effectively is the same?

oracle/types/market.go
```diff
@@ -13,10 +13,6 @@ func ProviderTickersFromMarketMap(
 	name string,
 	marketMap mmtypes.MarketMap,
 ) ([]ProviderTicker, error) {
-	if err := marketMap.ValidateBasic(); err != nil {
```
Eric-Warehime, line 16: Where does this get run now?

aljo242: This is always already run when the market map is read in

aljo242: https://github.com/skip-mev/slinky/blob/02f198691eadeb1cb1289d272989823e4c1549cc/x/marketmap/types/utils.go#L10

when read in from file

aljo242: https://github.com/skip-mev/slinky/blob/02f198691eadeb1cb1289d272989823e4c1549cc/providers/apis/marketmap/api_handler.go#L93

when read by provider

Eric-Warehime (approved):

providers/apis/dydx/multi_market_map_fetcher_test.go
```diff
@@ -96,6 +96,7 @@ func TestDYDXMultiMarketMapFetcher(t *testing.T) {
 									CurrencyPair:     slinkytypes.NewCurrencyPair("BTC", "USD"),
 									Decimals:         8,
 									MinProviderCount: 1,
+									Enabled:          true,
```
davidterpay, line 99: Does the test case below also have to have it be enabled?

aljo242: yeah, it should I guess - we are only checking that a field exists under that key.

not any of the fields themselves like `Enabled`

davidterpay (approved):

PR #400, opened 2024-05-07
chore: clean out `core/oracle.json`

aljo242:


davidterpay:
Can you add a label to run this with a VPN and the core market config?

PR #402, opened 2024-05-07
feat: use marketmap for deploys

aljo242:


contrib/images/slinky.sidecar.dev.Dockerfile
```diff
@@ -17,4 +17,4 @@ COPY --from=builder /src/slinky/build/* /usr/local/bin/
 RUN apt-get update && apt-get install ca-certificates -y
 
 WORKDIR /usr/local/bin/
-ENTRYPOINT ["slinky", "--oracle-config-path", "/oracle/oracle.json", "--market-config-path", "/oracle/market.json"]
+ENTRYPOINT ["slinky", "--oracle-config-path", "/oracle/oracle.json"]
```
davidterpay: Do you think it would be helpful to add the tag for update-market-config-path to write the market map when we get one?

contrib/images/slinky.sidecar.e2e.Dockerfile
```diff
@@ -20,4 +20,4 @@ COPY --from=builder /src/slinky/config/local /etc/slinky/default_config
 RUN apt-get update && apt-get install ca-certificates -y
 
 WORKDIR /usr/local/bin/
-ENTRYPOINT ["slinky", "--oracle-config-path", "/oracle/oracle.json", "--market-config-path", "/oracle/market.json"]
+ENTRYPOINT ["slinky", "--oracle-config-path", "/oracle/oracle.json"]
```
davidterpay: ditto above

contrib/images/slinky.sidecar.prod.Dockerfile
```diff
@@ -16,4 +16,4 @@ EXPOSE 8002
 COPY --from=builder /src/slinky/build/* /usr/local/bin/
 
 WORKDIR /usr/local/bin/
-ENTRYPOINT ["slinky", "--oracle-config-path", "/oracle/oracle.json", "--market-config-path", "/oracle/market.json"]
+ENTRYPOINT ["slinky", "--oracle-config-path", "/oracle/oracle.json"]
```
davidterpay: ditto above

contrib/images/slinky.sidecar.dev.Dockerfile
davidterpay: can probably write to the same directory

scripts/genesis.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+ set -e
```
davidterpay: i have no idea how to review this lmao

davidterpay (approved):

scripts/genesis.sh
aljo242: @Zygimantass was my sherpa here

providers/factories/oracle/marketmap.go
```diff
@@ -59,7 +62,18 @@ func MarketMapProviderFactory(
 		)
 		ids = []types.Chain{{ChainID: dydx.ChainID}}
 	default:
-		apiDataHandler, err = marketmap.NewAPIHandler(cfg.API)
+		var client mmtypes.QueryClient
+		client, err = marketmap.NewGRPCClient(cfg.API)
```
Eric-Warehime, line 66: Why do we want to construct the grpc client and pass it in instead of passing in the API config and constructing the client in the provider constructor?

oracle/orchestrator/market_mapper.go
```diff
@@ -31,16 +31,23 @@ func (o *ProviderOrchestrator) listenForMarketMapUpdates(ctx context.Context) {
 			// Fetch the latest market map.
 			response := mmProvider.GetData()
 			if response == nil {
-				o.logger.Debug("market map provider returned nil response")
+				o.logger.Info("market map provider returned nil response")
 				continue
 			}
 
+			o.logger.Info("market map provider received update")
```
Eric-Warehime: This is going to repeat every N seconds, but doesn't really give us any further insight into how the system is running. Overall I'd say it's not a useful log.

config/core/oracle.json
```diff
@@ -417,6 +417,22 @@
...
+    {
+      "name": "marketmap_api",
+      "api": {
+        "enabled": true,
+        "timeout": 20000000000,
+        "interval": 10000000000,
+        "reconnectTimeout": 2000000000,
+        "maxQueries": 1,
+        "atomic": true,
+        "url": "blockchain:9090",
```
Eric-Warehime: This works in our docker compose, but I think it's a weird default to have otherwise. It won't work if people are running the sidecar via systemd.

providers/factories/oracle/marketmap.go
davidterpay: Primarily for testing, the mm provider expects the query client interface.

oracle/orchestrator/market_mapper.go
davidterpay: we can probably only log on actual updates.

davidterpay: and move this to debug.

providers/factories/oracle/marketmap.go
aljo242: this is from @davidterpay PR.  Looks like this could be refactored

davidterpay: idt we should refactor, in the case where want to do something like a multi-client connection for a super-sidecar we can repurpose the the same interface. Additionally, the interface makes it much easier to test. @Eric-Warehime do you feel like this a blocker?

davidterpay: we mock out the client connection for essentially all of our non-rest providers (uniswap and raydium).

Makefile
```diff
@@ -27,6 +27,7 @@ SOLANA_NODE_ENDPOINT ?= https://api.devnet.solana.com
 ORACLE_GROUP ?= core
 GENESIS_MARKETS ?=  $(CONFIG_DIR)/$(ORACLE_GROUP)/market.json
 MARKETS := $(shell cat ${GENESIS_MARKETS})
+DEV_COMPOSE := $(CURDIR)/contrib/docker-compose.yml
```
davidterpay: Can we move this to a compose directory and have a docker-compose-dev.yml file.

davidterpay: This will be useful when the validators port over their compose into our repo.

metrics.md
```diff
@@ -25,9 +25,9 @@ The slinky repo contains a sidecar Grafana dashboard that can be utilized across
...
+To set up the sidecar Grafana dashboard, you will need to import the [`side-car-dashboard.json`](contrib/grafana/grafana/provisioning/dashboards/side-car-dashboard.json) file. This file contains the JSON representation of the dashboard. You can import the dashboard by following these steps:
 
-1. Add the side car prometheus data source to your Grafana instance. This can be done by adding a new data source in your datasources directory. An example can be found [here](./grafana/provisioning/datasources/prometheus.yml). Note that the `url` field should be the URL of your sidecar Prometheus instance.
+1. Add the side car prometheus data source to your Grafana instance. This can be done by adding a new data source in your datasources directory. An example can be found [here](contrib/grafana/grafana/provisioning/datasources/prometheus.yml). Note that the `url` field should be the URL of your sidecar Prometheus instance.
```
davidterpay: why are we moving the grafana dashboards into the contrib directory? IDT this is necessary as validators can currently use the dashboard if they want it.

davidterpay: IMO i think we should only move the top level compose.

davidterpay: its standard to have grafana follow the file structure currently in the codebase.

contrib/docker-compose.yml
```diff
@@ -3,18 +3,38 @@ networks:
...
+      dockerfile: ./contrib/images/slinky.local.Dockerfile
+    volumes:
+      - ../config/${ORACLE_GROUP}/market.json:/oracle/market.json
+    ports:
+      - "26656:26656"
+      - "26657:26657"
+      - "26660:26660" # chain side metrics
+      - "1317:1317" # used to query the blockchain
+      - "7171:7171"
+      - "26655:26655"
```
davidterpay: You have to also export the 9090 port no?

providers/factories/oracle/marketmap.go
davidterpay: oh i see what you mean now, I can just create a wrapper around the logic and push to the PR.

Eric-Warehime: It's not blocking just a preference thing I suppose.

Eric-Warehime (approved):

PR #403, opened 2024-05-07
feat: Hashing IDs Currency Pair Strategy

davidterpay:
- implementing a new currency pair strategy that will broadcast (sha256 hash of each CP, raw price) per vote.
- implements the same caching strategy as other strategies so there should be a similar amount of reads as the delta strategy.
- amount of overhead in terms of VE size should be constant but slightly larger than default.


beer-1 (approved):
LGTM 

overhead slightly lager, but I think this way is also good. Thanks for update!

abci/strategies/currencypair/hash.go
```diff
@@ -0,0 +1,98 @@
...
+		s.previousHeight = height
+	}
+
+	cp, found := s.idCache[id]
+	if found {
+		return cp, nil
+	}
+
+	// if the currency pair is not found in the cache, attempt to retrieve it from
+	// the x/oracle state by populating the cache with all currency pairs. This
```
Eric-Warehime, line 68: Is it a problem in terms of CP removal that we never evict entries from this cache?

Eric-Warehime (approved):

abci/strategies/currencypair/hash.go
davidterpay: This should be evicted once per height.

aljo242 (approved):

abci/strategies/currencypair/hash.go
```diff
@@ -0,0 +1,98 @@
...
+	if !found {
+		return slinkytypes.CurrencyPair{}, fmt.Errorf("currency pair with sha256 hashed ID %d not found in x/oracle state", id)
+	}
+
+	return cp, nil
+}
+
+// CurrencyPairToHashID returns the ID of the given currency pair, by taking the
+// sha256 hash of the currency pair.
+func CurrencyPairToHashID(currencyPair string) (uint64, error) { //nolint
```
g3ndo9kar9, line 90: We're truncating the hash here? Is that ok?

g3ndo9kar9 (commented):
Can we use this with the delta encoding strategy?

davidterpay:
> Can we use this with the delta encoding strategy?

roll ups need the raw price. otherwise they would need something like ICQ which is not part of the design.

abci/strategies/currencypair/hash.go
davidterpay: this is how it was implemented in the forked version, so i just modeled that. assuming it functions as desired.

g3ndo9kar9 (approved):
lgtm

PR #407, opened 2024-05-08
chore: Fix MarketMap Provider to use gRPC

davidterpay:


providers/apis/marketmap/utils.go
```diff
@@ -20,5 +20,5 @@ var DefaultAPIConfig = config.APIConfig{
 	Interval:         1 * time.Second,
 	ReconnectTimeout: 2000 * time.Millisecond,
 	MaxQueries:       1,
-	URL:              "http://localhost:1317/slinky/marketmap/v1/marketmap",
+	URL:              "http://localhost:9090",
```
aljo242, line 23: don't we need the full url for the marketmap?

aljo242 (approved):

PR #408, opened 2024-05-08
feat: Add exact HTTP codes for metrics collection

davidterpay:


providers/base/api/metrics/api_query_handler.go
```diff
@@ -70,7 +74,12 @@ func NewAPIMetrics() APIMetrics {
 		apiHTTPStatusCodePerProvider: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: oraclemetrics.OracleSubsystem,
 			Name:      "api_http_status_code",
-			Help:      "Number of API provider responses by status code.",
+			Help:      "Number of API provider responses by status code grouped by category (2XX, 3XX, etc.).",
+		}, []string{providermetrics.ProviderLabel, StatusCodeLabel}),
+		apiHTTPStatusCodeExactPerProvider: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: oraclemetrics.OracleSubsystem,
+			Name:      "api_http_status_code_granular",
+			Help:      "Number of API provider responses by status code with granularity on the exact code.",
```
davidterpay: adding a new metric just so things are backwards compatible and no promql wizardy is required to group status codes (since they are strings and would require regex).

grafana/provisioning/dashboards/side-car-dashboard.json
```diff
@@ -377,7 +377,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sum by (id) \n(rate(side_car_health_check_ticker_updates_total{job=\"${job}\"}[$__range]))\n/\nscalar(rate(side_car_health_check_system_updates_total{job=\"${job}\"}[$__range]))",
+          "expr": "sum by (id) \n(rate(side_car_health_check_ticker_updates_total{job=~\"${job}\"}[$__range]))\n/\nscalar(rate(side_car_health_check_system_updates_total{job=~\"${job}\"}[$__range]))",
```
davidterpay, line 379: Updating the dashboard as well for testing

davidterpay:
Added a preview environment so we can check out the new codes

providers/base/api/metrics/api_query_handler.go
Zygimantass: instead of a new metric, could we just add a new label?

Zygimantass: status_code and exact_status_code or smth?

davidterpay: fixed

davidterpay: good shout

Eric-Warehime: No, I don't think we should have a new metric at all. We should just have `api_http_status_code` with labels `status_code` and `status_code_exact`. So we just have one metric and you can use the `status_code` label to aggregate across `2XX`, `4XX`, etc. and the `status_code_exact` label to see if we're specifically hitting `404`s or `429`s.

davidterpay: tested with existing dashboards and it works with the single metric with the added label

Eric-Warehime (approved):

PR #413, opened 2024-05-09
feat: basic upgrade playbook

aljo242:
closes BLO-1225


upgrades/core/upgrades.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("invalid params: %w", err)
+	}
+
+	return marketmapKeeper.SetParams(ctx, params)
+}
+
+func setMarketState(ctx sdk.Context, mmKeeper *marketmapkeeper.Keeper, markets marketmaptypes.Markets) error {
+	// markets, err := marketmaptypes.ReadMarketsFromFile("markets.json")
+	//if err != nil {
+	//	return err
```
Eric-Warehime: Needed?

upgrades/core/markets.json
```diff
@@ -0,0 +1,4056 @@
+[
```
davidterpay, line 1: Is there a reason we need another core market json? We should ideally just have a single central one.

davidterpay: gonna be a footgun trying to maintain the same file in different places.

x/marketmap/types/testdata/marketmap.json
```diff
@@ -0,0 +1,4058 @@
+{
```
davidterpay, line 1: same applies here

upgrades/core/markets.json
aljo242: this is the markets as a list - i can just generate them on the fly

PR #420, opened 2024-05-10
test(`integration`): sidecar uses `x/marketmap`

aljo242:


abci/strategies/currencypair/hash_test.go
```diff
@@ -4,11 +4,12 @@ import (
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/mock"
```
g3ndo9kar9, line 7: is this from your local linter?

g3ndo9kar9 (approved):
lgtm

abci/strategies/currencypair/hash_test.go
aljo242: just ran `make format` before committing

PR #421, opened 2024-05-10
feat: Rotating Logs that can be Stored Locally by Default

davidterpay:
By default, sidecar logs will be written to `sidecar.log` with a max size of 100mb, 3 days of logs, 0 backups (i dont want to ballon this in case validators restart often), and info level logging (we can change once we reduce the amount of debug logs being written).


cmd/slinky/main.go
```diff
@@ -90,11 +91,18 @@ func init() {
 		"info",
 		"Log level (debug, info, warn, error, dpanic, panic, fatal).",
 	)
+	rootCmd.Flags().StringVarP(
+		&fileLogLevel,
+		"file-log-level",
+		"",
+		"info",
```
davidterpay, line 102: can move this to debug once we purge and bubble up some of the logs.

aljo242: make linear task for this?

pkg/log/zap.go
```diff
@@ -1,77 +1,101 @@
...
-	// outputPaths is the list of paths to write logs to. By default, logs are
-	// written to stderr.
-	outputPaths := []string{"stderr"}
-	if config.WriteTo != "" {
-		outputPaths = append(outputPaths, config.WriteTo)
+	// Setup the primary output to always include os.Stderr
+	stderrSyncer := zapcore.Lock(os.Stderr)
+
+	var fileCore zapcore.Core
+	if config.WriteTo != "" && config.WriteTo != "stderr" {
```
aljo242: let's make "stderr" a `const`

aljo242 (approved):
lgtm w comment

PR #425, opened 2024-05-13
feat: Sensible Internal Defaults [BLO-1251]

Eric-Warehime:
## In This PR
- I update the `slinky` command to ingest configs in the following way
- Running `./slinky` will reference a default config as defined in `cmd/constants/providers.go`
- Operators can manually specify overrides via environment variables as follows
```
SLINKY_CONFIG_PROVIDERS_MARKETMAP_API_API_URL=abc
SLINKY_CONFIG_PROVIDERS_DYDX_API_API_URL=abc
SLINKY_CONFIG_PROVIDERS_RAYDIUM_API_API_ENDPOINTS_0_AUTHENTICATION_APIKEYHEADER=abc
```
- If the `--oracle-config` command is specified with a path to an `oracle.json`, the binary interprets a sparsely defined oracle json, interpreting missing values as retaining defaults
- The default market-map provider used is the `marketmap_api`, the `--marketmap-provider` flag allows operators to identify a market-map provider to use (by name)


cmd/slinky/config.go
```diff
@@ -0,0 +1,147 @@
...
+	DefaultPrometheusServerAddress = "0.0.0.0:8002"
+	// DefaultMetricsEnabled is the default value for enabling prometheus metrics in slinky.
+	DefaultMetricsEnabled = true
+	// DefaultHost is the default for the slinky oracle server host.
+	DefaultHost = "0.0.0.0"
+	// DefaultPort is the default for the slinky oracle server port.
+	DefaultPort = "8080"
+)
+
+type OracleConfig struct {
```
g3ndo9kar9, line 54: Is this idea for us to have the `LegacyOracleConfig` be a sidecar thing, and this `OracleConfig` be a separate config?

g3ndo9kar9: If so, I'd prefer that we canonicalize this version of the config + handle the adaptation between legacy config / this one during reading the legacy config from file?

cmd/slinky/config.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func GetLegacyOracleConfig(path string) (config.OracleConfig, error) {
+	SetDefaults()
+	var oracleCfg OracleConfig
+	var err error
+	if path != "" {
+		oracleCfg, err = ReadOracleConfigFromFile(path)
+	} else {
+		err = viper.Unmarshal(&oracleCfg)
```
g3ndo9kar9: This always unmarshals to a default oracle config? Like if no path specified only happens when I want no modifications?

g3ndo9kar9: so essentially, this is basically viper starts w/ a default config. If an oracle path is specified, viper reads in that + mutates fields specified. Then we unmarshal / validate

PR #431, opened 2024-05-14
fix(providers/raydium): increase raydium provider timeout

g3ndo9kar9:
## In This PR
- I increase the raydium provider's timeout from 500ms to 1 second


aljo242 (approved):

davidterpay (approved):

providers/apis/defi/raydium/price_fetcher.go
```diff
@@ -70,7 +70,14 @@ func NewAPIPriceFetcher(
 				client,
 			))
 		} else {
-			config.URL = config.Endpoints[0].URL
+			client, err := solanaClientFromEndpoint(config.Endpoints[0])
```
Eric-Warehime: This prevents people from using URL. Should probably fail validation if they only provide a URL then?

Eric-Warehime (approved):

PR #441, opened 2024-05-21
feat: Update slinky driver to use default configs [BLO-1251]

g3ndo9kar9:
## In This PR
- I update the `slinky` command to ingest configs in the following way
- Running `./slinky` will reference a default config as defined in `cmd/constants/providers.go`
- Operators can manually specify overrides via environment variables as follows
```
SLINKY_CONFIG_PROVIDERS_MARKETMAP_API_API_URL=abc
SLINKY_CONFIG_PROVIDERS_DYDX_API_API_URL=abc
SLINKY_CONFIG_PROVIDERS_RAYDIUM_API_API_ENDPOINTS_0_AUTHENTICATION_APIKEYHEADER=abc
```
- If the `--oracle-config` command is specified with a path to an `oracle.json`, the binary interprets a sparsely defined oracle json, interpreting missing values as retaining defaults
- The default market-map provider used is the `marketmap_api`, the `--marketmap-provider` flag allows operators to identify a market-map provider to use (by name)


cmd/slinky/config.go
```diff
@@ -0,0 +1,274 @@
...
+		Host:           c.Host,
+		Port:           c.Port,
+	}
+}
+
+func SetDefaults() {
+	setViperDefaultsForDataStructure("", DefaultOracleConfig())
+}
+
+func setViperDefaultsForDataStructure(keyPrefix string, config interface{}) {
```
aljo242: Can we test this func?

aljo242 (commented):
overall lgtm - might need some testing around the new functions

cmd/slinky/main.go
```diff
@@ -66,10 +69,24 @@ var (
...
+		"",
+		marketmap.Name,
+		"MarketMap provider to use (marketmap_api, dydx_api, dydx_research_api).",
+	)
+	rootCmd.Flags().StringVarP(
+		&legacyOracleCfgPath,
 		"oracle-config-path",
 		"",
-		"oracle.json",
+		"",
```
Eric-Warehime, line 87: Changing the name of the flag will fully break people and even if we keep the flag name the same, changing the default will probably break some people. I think it's fine as long as we call everything out clearly in our docs. We can maybe have a section which details `0.4.X` config and its deprecation/migration path going forward. 

It seems like we want to roll this into `1.0.0` as a breaking change and maybe fully remove the old config in `1.1.X`? Let me know how you're thinking about this as a user experience.

Eric-Warehime: We probably want to use this https://pkg.go.dev/github.com/spf13/pflag#FlagSet.MarkDeprecated

cmd/constants/providers.go
```diff
@@ -0,0 +1,142 @@
...
+			Type:      types.ConfigType,
+		},
+		{
+			Name:      okx.Name,
+			WebSocket: okx.DefaultWebSocketConfig,
+			Type:      types.ConfigType,
+		},
+	}
+
+	MarketMapProviders = []config.ProviderConfig{
```
Eric-Warehime: `AlternativeMarketMapProviders` since the default one is already in the main config?

cmd/slinky/main.go
```diff
@@ -66,10 +69,24 @@ var (
 
 func init() {
 	rootCmd.Flags().StringVarP(
-		&oracleCfgPath,
+		&marketMapProvider,
+		"marketmap-provider",
+		"",
+		marketmap.Name,
+		"MarketMap provider to use (marketmap_api, dydx_api, dydx_research_api).",
```
Eric-Warehime: Do you think it's a good idea to leave `dydx_research_api` undocumented? Seems like something we don't really want anyone else to use.

g3ndo9kar9:
Update to slinky docs for neutron [here](https://github.com/skip-mev/docs/pull/93). This is the instructions we'll give to validators for running slinky using the new `slinky` entrypoint

cmd/slinky/config.go
```diff
@@ -0,0 +1,274 @@
...
+	MaxPriceAge time.Duration `json:"maxPriceAge"`
+
+	// Providers is the map of provider names to providers that the oracle will fetch prices from.
+	Providers map[string]config.ProviderConfig `json:"providers"`
+
+	// Production specifies whether the oracle is running in production mode. This is used to
+	// determine whether the oracle should be run in debug mode or not.
+	//
+	// Deprecated: This field is no longer used.
+	Production bool `json:"production"`
```
Eric-Warehime: This is effectively a new config which is incompatible with the existing config. We should just remove this in this PR.

cmd/slinky/config.go
```diff
@@ -0,0 +1,274 @@
...
+			}
+		default:
+			viper.SetDefault(fullKey, field.Interface())
+		}
+	}
+
+	// set the environment prefix
+	viper.SetEnvPrefix("SLINKY_CONFIG")
+	viper.AutomaticEnv()
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
```
Eric-Warehime: Curious why you're using `_` instead of `.` for env variables.

Eric-Warehime:
> Update to slinky docs for neutron [here](https://github.com/skip-mev/docs/pull/93). This is the instructions we'll give to validators for running slinky using the new `slinky` entrypoint

Your docs mention this as `0.4.2`. This change will fully break everyone so I don't really think we can do that. 

Edit: We're already at 0.4.4 anyways.

cmd/slinky/config.go
g3ndo9kar9: . doesn't work as an env variable for some reason?

cmd/slinky/main.go
g3ndo9kar9: THat makes sense

aljo242:
@nivasan1 e2e is failing due to something with the `oracle.json` cfg file

cmd/slinky/main.go
```diff
@@ -185,9 +211,36 @@ func runOracle() error {
...
+	logCfg.Compress = !disableCompressLogs
+
+	// Build logger.
+	logger := log.NewLogger(logCfg)
+	defer logger.Sync()
+
+	var cfg config.OracleConfig
+	var err error
+	if legacyOracleCfgPath != "" && !useDefaults {
+		logger.Info("The --oracle-config-path flag is deprecated and will be removed in a future release. Please use --default-config --oracle-config instead.")
```
Eric-Warehime: I don't really like having `--default-config` be permanent. I think I'd rather just have
`slinky --oracle-config-path ./foobar.json` parse the existing config
`slinky --oracle-config ./foobar.json` parse the new config
`slinky` attempts to parse ./oracle.json as the old config and if not present launches with the defaults.

What do you think? That would make it cleaner when we just remove the old config in a breaking release.

cmd/constants/providers.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/skip-mev/slinky/providers/websockets/huobi"
+	"github.com/skip-mev/slinky/providers/websockets/kraken"
+	"github.com/skip-mev/slinky/providers/websockets/kucoin"
+	"github.com/skip-mev/slinky/providers/websockets/mexc"
+	"github.com/skip-mev/slinky/providers/websockets/okx"
+	mmtypes "github.com/skip-mev/slinky/service/clients/marketmap/types"
+)
+
+var (
+	Providers = []config.ProviderConfig{
```
davidterpay, line 31: small nit: can we sort these by api, defi, ws, and marketmap. sorted alphabetically in each sub-group.

Eric-Warehime (approved):

cmd/slinky/main.go
```diff
@@ -64,12 +68,30 @@ var (
 	disableRotatingLogs bool
 )
 
+const (
+	DefaultLegacyConfigPath = "./oracle.json"
+)
```
davidterpay, line 73: Where is this defined atm? Im assuming you want to have this be in `./config`?

cmd/slinky/main.go
```diff
@@ -323,9 +355,41 @@ func runOracle() error {
...
+// based on the provided flags.
+func useLegacyOracleConfig(logger *zap.Logger) (string, bool) {
+	// if --oracle-config has been specified, use that
+	if oracleCfgPath != "" {
+		return oracleCfgPath, false
+	}
+
+	// if a value is provided for the --oracle-config-path flag, use it
+	if legacyOracleCfgPath != "" {
+		logger.Info("DEPRECATION WARNING:: The --oracle-config-path flag is deprecated and will be removed in v1.0.0. Please use --default-config --oracle-config instead.")
```
davidterpay: nit: would move these logs to `.Warn`

cmd/slinky/config/config.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// oracleConfigFromViper unmarshals an oracle config from viper, validates it, and returns it.
+func oracleConfigFromViper() (OracleConfig, error) {
+	var cfg OracleConfig
+	if err := viper.Unmarshal(&cfg); err != nil {
+		return OracleConfig{}, err
+	}
+
+	// for each api-provider, we'll have to manually fill the endpoints
```
davidterpay, line 217: This is only allowing us to use env variables to modify the api endpoints? Do we want to be able to modify things like host, ws, and other?

g3ndo9kar9: discussed offline

davidterpay (approved):

PR #444, opened 2024-05-22
fix: add user-agent header to outbound requests to solana ndoes

g3ndo9kar9:
## In This PR
- I adjust the raydium provider to attach a user-agent header to all outbound requests.
- The user-agent header is mapped to the current slinky semantic version in user (in accordance with the `cmd.Build` value)
- I've tested this locall
Screen Shot 2024-05-22 at 11 02 39 AM
y via nc -l

### DIFF
--- a/pkg/http/round_tripper_with_headers_test.go
+++ b/pkg/http/round_tripper_with_headers_test.go
@@ -19,7 +19,7 @@ func TestRoundTripperWithHeaders(t *testing.T) {
 		expectedHeaderFields: expectedHeaderFields,
 	}
 
-	rtWithHeaders := slinkyhttp.NewRoundTripperWithHeaders(expectedHeaderFields, rt)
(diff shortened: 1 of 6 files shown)


aljo242 (approved):

pkg/http/round_tripper_with_headers_test.go
```diff
@@ -19,7 +19,7 @@ func TestRoundTripperWithHeaders(t *testing.T) {
 		expectedHeaderFields: expectedHeaderFields,
 	}
 
-	rtWithHeaders := slinkyhttp.NewRoundTripperWithHeaders(expectedHeaderFields, rt)
+	rtWithHeaders := slinkyhttp.NewRoundTripperWithHeaders(rt, slinkyhttp.WithAuthentication("X-Api-Key", "test"))
```
Eric-Warehime, line 22: Can we just add the expected user agent header here? Or is it difficult to test because it includes the commit sha on non-releases?

Eric-Warehime (approved):

PR #448, opened 2024-05-23
feat: Swap REST for WS for Coinbase in Core

davidterpay:


davidterpay:
reference grafana deployment here: https://skipprotocol.grafana.net/d/cdca49ue665mocd/slinky-sidecar-dashboard?orgId=1&var-DS_PROMETHEUS=default&var-job=slinky-preview-pr-448-sidecar-0&var-id=All&var-provider=All&var-market_map_provider=dydx_api&from=1716573780112&to=1716574641473

goal here is that the uptime for coinbase_ws is 100%

oracle/orchestrator/market_mapper.go
```diff
@@ -41,8 +41,8 @@ func (o *ProviderOrchestrator) listenForMarketMapUpdates(ctx context.Context) {
 				continue
 			}
 
-			if result.Code.Error() != nil {
-				o.logger.Info("market map provider response error", zap.Error(result.Code.Error()))
+			if result.ErrorCode.Error() != nil {
```
aljo242: what is this change?

providers/apis/dydx/helper_test.go
```diff
@@ -94,7 +94,7 @@ var convertedResponse = mmtypes.MarketMapResponse{
 						NormalizeByPair: &usdtusd,
 					},
 					{
-						Name:           coinbaseapi.Name,
+						Name:           coinbasews.Name,
```
aljo242, line 97: Can we remove the API provider entirely now?

since it does not work and we no longer support

providers/apis/dydx/parse_test.go
```diff
@@ -9,11 +9,11 @@ import (
 
 	"github.com/skip-mev/slinky/oracle/constants"
 	slinkytypes "github.com/skip-mev/slinky/pkg/types"
-	coinbaseapi "github.com/skip-mev/slinky/providers/apis/coinbase"
 	"github.com/skip-mev/slinky/providers/apis/defi/raydium"
 	"github.com/skip-mev/slinky/providers/apis/defi/uniswapv3"
 	"github.com/skip-mev/slinky/providers/apis/dydx"
 	dydxtypes "github.com/skip-mev/slinky/providers/apis/dydx/types"
+	coinbasews "github.com/skip-mev/slinky/providers/websockets/coinbase"
```
aljo242, line 16: can we just use `coinbase`, as it is the canonical implementation of the provider ?

providers/apis/dydx/parse.go
```diff
@@ -19,6 +18,7 @@ import (
 	"github.com/skip-mev/slinky/providers/websockets/bitfinex"
 	"github.com/skip-mev/slinky/providers/websockets/bitstamp"
 	"github.com/skip-mev/slinky/providers/websockets/bybit"
+	"github.com/skip-mev/slinky/providers/websockets/coinbase"
```
davidterpay, line 21: Replacing REST with WS

aljo242 (approved):

oracle/orchestrator/market_mapper.go
davidterpay: I am swapping out the error code out of the resolved result since it should not self-contain an error, instead I mark it with a response code which can give us insight into the type of update (similar to how http has different 2XXs).

providers/apis/dydx/helper_test.go
davidterpay: wym remove? We should still keep the implementation, but it's not being used in our configs anymore.

PR #454, opened 2024-05-28
feat: Binance WS Implementation

davidterpay:


oracle/types/provider.go
```diff
@@ -82,5 +83,7 @@ func (t *ProviderTickers) Add(ticker ProviderTicker) {
 	t.mut.Lock()
 	defer t.mut.Unlock()
 
+	t.cache[strings.ToLower(ticker.GetOffChainTicker())] = ticker
 	t.cache[ticker.GetOffChainTicker()] = ticker
+	t.cache[strings.ToUpper(ticker.GetOffChainTicker())] = ticker
```
davidterpay, line 88: Adding this since some times the subscription is lower case but response is upper case. Should be a safe operation.

Eric-Warehime: Why do we need all 3 of these instead of using strictly upper or strictly lower?

davidterpay: brevity mostly, ideally we just use some type of regex.

providers/websockets/binance/README.md
```diff
@@ -0,0 +1,16 @@
+# Binance Provider
+
+## Overview
+
+The Binance provider is used to fetch the ticker price from the [Binance websocket API](https://developers.binance.com/docs/binance-spot-api-docs/web-socket-streams). A single connection is only valid for 24 hours; after that, a new connection must be established. The Websocket server will send a ping frame every 3 minutes. If the client does not receive a pong frame within 10 minutes, the connection will be closed. Note that all symbols are in lowercase.
+
+The WebSocket connection have a limit of 5 incoming messages per second. A message is considered:
```
technicallyty:
```suggestion
The WebSocket connection has a limit of 5 incoming messages per second. A message is considered:
```

Eric-Warehime (approved):

oracle/types/provider.go
Eric-Warehime: I don't think this makes sense. Ideally something like this works:

Adding to cache:
```
t.cache[strings.ToUpper(ticker.GetOffChainTicker())] = ticker
```
Retrieving from cache:
```
return t.cache[strings.ToUpper(ticker.GetOffChainTicker())]
```

Why do we need a regex ever?

providers/websockets/binance/messages.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// GenerateID generates a random ID for the message.
+func (h *WebSocketHandler) GenerateID() int64 {
+	for {
+		id := rand.Int63() + 1
+		if _, ok := h.messageIDs[id]; !ok {
+			return id
+		}
+	}
```
technicallyty: alternatively you could store an int64 on the handler, and bump it everytime you need a new ID? might be a bit simpler, but also might be optimization mogging you with this comment lol

providers/websockets/binance/ws_data_handler.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Copy is used to create a copy of the WebSocketHandler.
+func (h *WebSocketHandler) Copy() types.PriceWebSocketDataHandler {
+	return &WebSocketHandler{
+		logger:     h.logger,
+		ws:         h.ws,
+		cache:      types.NewProviderTickers(),
+		messageIDs: make(map[int64][]string),
+	}
+}
```
technicallyty: could you just call `NewWebSocketHandler`? might be safer if the constructor changes overtime, and one forgets to update this manual constructor here

providers/websockets/binance/messages.go
davidterpay: should have done that to begin with, this for loop scares me anyways lol

oracle/types/provider.go
davidterpay: some providers dont specify whether we retrieve responses in the same casing as the request, which could lead to total liveness failure on a given provider. regex is probably overkill but if the offchain ticker happens to be upper case and responses are lower case we would effectively not be able to parse out the responses.

davidterpay: We could change the API to the `ProviderTickers` func but that would be a much larger change.

providers/websockets/binance/ws_data_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	if err := ws.ValidateBasic(); err != nil {
+		return nil, fmt.Errorf("invalid websocket config for %s: %w", Name, err)
+	}
+
+	return &WebSocketHandler{
+		logger:     logger,
+		ws:         ws,
+		cache:      types.NewProviderTickers(),
+		messageIDs: make(map[int64][]string),
+		nextID:     rand.Int63() + 1,
```
technicallyty, line 53: any reason not to just start it @ 1?

davidterpay: since the connection can be multi-plexed we might duplicate the ID which might affect server side request handling.

providers/websockets/binance/ws_data_handler.go
davidterpay: we might to deal with a panic if for whatever reason the `NewWebSocketHandler` returns an error. We've also just stuck to reconstructing the handler in the other providers this way so probably just prefer sticking to this for the time being.

davidterpay: at some point we talked about creating a parent class for this exact reason, probably not high prio for the time being.

PR #464, opened 2024-05-30
fix: deploy on main as well

Zygimantass:


davidterpay:
This gets us deployments in k8s everytime we push to main?

aljo242 (approved):

Zygimantass:
@davidterpay yessir, but only on the slinky-dev instance (not the customer deployments)

davidterpay (approved):

PR #466, opened 2024-05-30
chore: Add Core + Raydium Markets into Exportable Go structs

davidterpay:


cmd/constants/markets.go
```diff
@@ -0,0 +1,5340 @@
...
+				  "Quote": "USD"
+				}
+			  },
+			  {
+				"name": "kraken_api",
+				"off_chain_ticker": "XXRPZUSD"
+			  },
+			  {
+				"name": "kucoin_ws",
+				"off_chain_ticker": "XRP-USDT",
```
aljo242, line 5133: Can we have these as native go structs rather than as json that we / unmarshal?

config/core/market.json
```diff
@@ -151,6 +151,14 @@
             "Base": "USDT",
             "Quote": "USD"
           }
+        },
```
aljo242: This file is being removed?

cmd/constants/markets.go
davidterpay: replicating the core market is gonna take me a fat minute. this effectively accomplishes the same thing

config/core/market.json
davidterpay: not in this PR but in a future one

aljo242 (approved):

PR #470, opened 2024-05-31
chore: Removing market json files in repo

davidterpay:
The build workflow now looks like:
1. Specify the markets you want to run (core, raydium, uniswap, etc.) as environment variables.
2. Merge into a single marketmap.
3. Write marketmap to genesis of app.
4. Compose will spin up the node along with sidecar pointing to the node.
5. Mog


Makefile
```diff
@@ -164,10 +133,15 @@ $(BUILD_TARGETS): $(BUILD_DIR)/
 $(BUILD_DIR)/:
 	@mkdir -p $(BUILD_DIR)/
 
-# build-configs builds a slinky simulation application binary in the build folder (/test/.slinkyd)
-build-configs:
+delete-configs:
```
aljo242, line 136: can you add `.PHONY`

aljo242 (approved):

scripts/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"flag"
+	"fmt"
+
+	"github.com/skip-mev/slinky/cmd/constants"
+	mmtypes "github.com/skip-mev/slinky/x/marketmap/types"
+)
+
+var (
+	useCore    = flag.Bool("use-core", false, "use core markets")
```
g3ndo9kar9, line 15: we only have raydium / core? Ig for dydx / research-markets we're j pointing at the external sources (dydx-nodes, research-json)

g3ndo9kar9 (approved):
lgtm

scripts/genesis.go
davidterpay: yea thats the plan

PR #482, opened 2024-06-04
feat: decouple ve handler from preblocker [BLO-1327]

g3ndo9kar9:
## In This PR
- I migrate the logic (previously in the `PreBlockHandler`) for unmarshalling an extended-commit, translating votes into aggregated prices + writing those prices to state to a separate interface, the `PriceApplier`
- I have both the `PreBlockHandler` and `VoteExtensionHandler` depend on this interface, instead of having the `VoteExtensionHandler` depend on a `PreBlockHandler`


tests/simapp/app.go
```diff
@@ -369,16 +370,34 @@ func NewSimApp(
...
-			compression.NewZLibCompressor(),
+		cps,
+		veCodec,
+		aggregator.NewOraclePriceApplier(
+			aggregator.NewDefaultVoteAggregator(
+				app.Logger(),
+				aggregatorFn,
+				// we need a separate price strategy here, so that we can optimistically apply the latest prices
+				// and extend our vote based on these prices
+				currencypair.NewDeltaCurrencyPairStrategy(app.OracleKeeper),
```
g3ndo9kar9, line 394: Currently, we need this `CurrencyPairStrategy` to be separate from the currency-pair strategy used by the `VoteExtensionHandler` bc of a nuance in the behavior of the `CurrencyPairStrategy`'s cache, where, since we're optimistically applying prices w/ height `req.Height`, the cache is used and caches a price per currency-pair before applying the new prices to state. Thus, when the `ExtendVoteHandler` is transforming prices from the oracle, it won't use the newly applied prices (if separate caches are not used).

Notice, this is currently also a problem if the same `CurrencyPairStrategy` is used in the `PreBlocker` and `VoteExtensionHandler`

aljo242 (approved):

Eric-Warehime (approved):
LGTM. Seems like the new interface would be more useful if it was passed in instead of constructing the default price applier inside the preblock handler. But since nobody is customizing it right now I think we can wait until we need to do that.

abci/strategies/aggregator/price_applier.go
```diff
@@ -0,0 +1,148 @@
...
+type PriceApplier interface {
+	// ApplyPricesFromVoteExtensions derives the aggregate prices per asset in accordance with the given
+	// vote extensions + VoteAggregator. If a price exists for an asset, it is written to state. The
+	// prices aggregated from vote-extensions are returned if no errors are encountered in execution,
+	// otherwise an error is returned + nil prices.
+	ApplyPricesFromVoteExtensions(ctx sdk.Context, req *cometabci.RequestFinalizeBlock) (map[slinkytypes.CurrencyPair]*big.Int, error)
+
+	// GetPriceForValidator gets the prices reported by a given validator. This method depends
+	// on the prices from the latest set of aggregated votes.
+	GetPriceForValidator(validator sdk.ConsAddress) map[slinkytypes.CurrencyPair]*big.Int
```
davidterpay:
```suggestion
	// GetPricesForValidator gets the prices reported by a given validator. This method depends
	// on the prices from the latest set of aggregated votes.
	GetPricesForValidator(validator sdk.ConsAddress) map[slinkytypes.CurrencyPair]*big.Int
```

abci/ve/vote_extension.go
```diff
@@ -56,7 +57,7 @@ func NewVoteExtensionHandler(
 	timeout time.Duration,
 	strategy currencypair.CurrencyPairStrategy,
 	codec compression.VoteExtensionCodec,
-	preBlocker sdk.PreBlocker,
+	priceAppllier aggregator.PriceApplier,
```
davidterpay: typo

davidterpay (approved):

PR #484, opened 2024-06-05
feat: add MarketMap query to oracle server

technicallyty:
- adds ~an http handler~ gRPC handler to the Oracle server to get latest marketmap from the ProviderOrchestrator. 
- adds option to Oracle constructor to pass in a function to get a marketmap
- extends Oracle interface with `GetMarketMap` method


tests/integration/go.mod
```diff
@@ -13,7 +13,7 @@ go 1.22.2
 require (
 	cosmossdk.io/math v1.3.0
 	github.com/cometbft/cometbft v0.38.7
-	github.com/cosmos/cosmos-sdk v0.50.6
+	github.com/cosmos/cosmos-sdk v0.50.7
```
technicallyty: GoLand will not stop updating this.... SEND HELP

oracle/oracle.go
```diff
@@ -259,3 +264,13 @@ func (o *OracleImpl) GetPrices() types.Prices {
...
+
+// GetMarketMap returns the current market map configuration.
+//
+// TODO(Tyler): is the empty map ok? maybe? maybe not. lord save us all.
+func (o *OracleImpl) GetMarketMap() marketmaptypes.MarketMap {
+	if o.marketMapGetter != nil {
+		return o.marketMapGetter()
+	}
+	return marketmaptypes.MarketMap{}
+}
```
technicallyty: if no option is passed in, ok to just empty return?

service/servers/oracle/server.go
```diff
@@ -195,6 +196,20 @@ func (os *OracleServer) Prices(ctx context.Context, req *types.QueryPricesReques
 	}
 }
 
+func (os *OracleServer) HandleGetMarketMap(w http.ResponseWriter, _ *http.Request) {
```
g3ndo9kar9: Instead of writing our own HTTP handler implementations, we usually define our services via [proto](https://github.com/skip-mev/slinky/blob/1657e81e45843d82d1831a18d2795d3fbae8b056/proto/slinky/service/v1/oracle.proto#L14), and use grpc's [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) functionality to spin up a proxy in front of our GRPC server to translate inbound REST requests. The advantage of doing this is that we can 1. define our req/res schema via protobuf, 2. version that schema better, 3. follow idioms defined in the codebase

TLDR: is that I recommend extending the service defined [here](https://github.com/skip-mev/slinky/blob/main/proto/slinky/service/v1/oracle.proto#L14) to include a MarketMap API, implement the service on the `OracleServer` as a GRPC route, and the REST route should be handled for us.

oracle/oracle.go
g3ndo9kar9: im cool w that

g3ndo9kar9: Alternatively, you can pass the `getMarketMapFn` closure as an argument to construct the `Oracle`, and fail the constructor if it's nil. Either way works IMO

tests/integration/go.mod
g3ndo9kar9: changelog is [here](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.50.7), doesn't look like anything super relevant to us, so this is safe+

g3ndo9kar9 (commented):
one comment, congrats on becoming an official oracle 👨‍🍳

service/servers/oracle/server.go
technicallyty: cc @Eric-Warehime had mentioned to just make it a http handler -- can switch to gRPC if thats desired! i agree it would make it more uniform w/ rest of code 👍🏻

oracle/oracle.go
technicallyty: totally -- just wasn't sure how delicate we wanted to be w/ breaking the constructor signature. 

or perhaps i should find a way to serve a default to the static config.. cause theres _always_ a marketmap, right? this endpoint is just to view it as it changes, so maybe serving whatever default (if there is one) would be nicer here?

cmd/slinky/main.go
```diff
@@ -289,7 +289,11 @@ func runOracle() error {
 	defer orch.Stop()
 
 	// Create the oracle and start the oracle server.
-	oracleOpts = append(oracleOpts, oracle.WithProviders(orch.GetPriceProviders()))
+	oracleOpts = append(
+		oracleOpts,
+		oracle.WithProviders(orch.GetPriceProviders()),
+		oracle.WithMarketMapGetter(orch.GetMarketMap),
```
aljo242, line 295: Why are we making this optional?

I think we should just pass a reference to the `orchestrator` into the `oracle` upon construction and use the `GetMarketMap` function directly.  There should only ever be one implementation of this `MarketMapGetter` function.

aljo242: This is kind of a broader discussion about how we construct the oracle in general.  Right now the oracle does not "own" or have a notion of the orchestrator, which seems confusing to me

technicallyty: yeah, that was confusing to me as well. it seemed the oracle needs to know about things the orchestrator is doing, so i thought id find a direct ref to it but found the options instead

i can address that in this PR if we'd like or can open a new linear ticket?

davidterpay: yea ideally we move away from the price oracle and move core logic into the orchestrator and make it the central info maintainer. Definitely not in scope for this PR, would open a separate issue.

davidterpay: Given we now have the RPC we can probably also remove the flag to write the marketmap locally (can also be a separate issue) in the orchestrator.

service/servers/oracle/server.go
davidterpay: also in favor of just adding an additional RPC method in proto.

cmd/slinky/main.go
aljo242: agreed - let's open a separate issue for refactoring the logic.


I think for the meantime, maybe we can just pass a ref to the orchestrator and always enable queries?

technicallyty: ticket [here](https://linear.app/skip/issue/BLO-1329/move-oracle-service-into-orchestrator), marking this as resolved

oracle/oracle.go
technicallyty: gonna mark this as resolved as this will be a tad different after BLO-1329

service/servers/oracle/server.go
technicallyty: this is now switched to gRPC, @davidterpay @nivasan1

aljo242 (approved):

Eric-Warehime (approved):
:shipit:

PR #491, opened 2024-06-05
feat: Adding Post Connection Timeout + Write Interval Timeouts

davidterpay:


oracle/config/websocket.go
```diff
@@ -70,6 +78,10 @@ type WebSocketConfig struct {
 	// to the websocket endpoint.
 	ReconnectionTimeout time.Duration `json:"reconnectionTimeout"`
 
+	// PostConnectionTimeout is the timeout for the provider to wait after a connection
+	// is established before sending messages.
+	PostConnectionTimeout time.Duration `json:"postConnectionTimeout"`
```
g3ndo9kar9, line 83: Timeout is a misleading name here? More like Wait, or smth

providers/base/websocket/handlers/ws_query_handler.go
```diff
@@ -159,6 +159,9 @@ func (h *WebSocketQueryHandlerImpl[K, V]) start() error {
 		return errors.ErrDialWithErr(err)
 	}
 
+	// Wait for the connection timeout before sending the initial payload(s).
+	time.Sleep(h.config.PostConnectionTimeout)
```
g3ndo9kar9, line 163: we can get rate-limited after the first connection?

g3ndo9kar9 (approved):
lgtm

providers/base/websocket/handlers/ws_query_handler.go
davidterpay: no messages should have been sent to the connection so should not be possible on init. it is possible if the connection is multi-plexed but as long as the multi-plexing is spaced out sufficiently we shouldnt see this issue. Ideally the time between spinning up concurrent connections is greater than the rate limit configs.

PR #495, opened 2024-06-06
refactor: parallelize MultiRPCClient's calls

technicallyty:
refactors the mutliclient to parallelize the client calls.

### DIFF
--- a/providers/apis/defi/ethmulticlient/multi_client.go
+++ b/providers/apis/defi/ethmulticlient/multi_client.go
@@ -95,7 +95,6 @@ func (m *MultiRPCClient) BatchCallContext(ctx context.Context, batchElems []rpc.
 		m.logger.Debug("BatchCallContext called with 0 elems")
 		return nil
 	}
-
 	// define a result struct that go routines will populate and append to a slice when they complete their request.
 	type result struct {
 		height  uint64
@@ -112,6 +111,7 @@ func (m *MultiRPCClient) BatchCallContext(ctx context.Context, batchElems []rpc.
 	for clientIdx, client := range m.clients {
 		wg.Add(1)
 		go func(i int) {
+			defer wg.Done()
 			url := m.api.Endpoints[i].URL
 
 			// append an eth_blockNumber call to the requests. we do this because we want the greatest height results only.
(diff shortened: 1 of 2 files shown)


providers/apis/defi/ethmulticlient/multi_client.go
```diff
@@ -87,77 +89,103 @@ func NewMultiRPCClientFromEndpoints(
...
-
-			maxHeight = newHeight
-			copy(batchElems, req[:blockNumReqIndex])
+	var maxHeightIndex int
+	for i, res := range results {
+		if res.height > maxHeight {
+			maxHeight = res.height
+			maxHeightIndex = i
 		}
 	}
```
technicallyty: i considered doing a buffered channel and doing height comparisons on the fly, but seemed potentially overkill. gathering the results then doing a once over seemed more straightforward. happy to be wrong here though. just lmk your thoughts <3

addtly, anyone know the typical amt of clients a multiclient would have?

providers/apis/defi/ethmulticlient/multi_client.go
```diff
@@ -87,77 +89,103 @@ func NewMultiRPCClientFromEndpoints(
...
-
+	// maxHeight being 0 means there were no results. something bad happened. return all the errors.
 	if maxHeight == 0 {
-		return errs
+		err := errors.Join(errs...)
+		if err != nil {
+			return err
+		}
+		// this should never happen... but who knows. maybe something terrible happened.
+		return errors.New("no errors were encountered, however no go routine was able to report a height")
```
technicallyty, line 189: not rly sure if this is worth adding but 🤷 can remove if its stupid lol

providers/apis/defi/ethmulticlient/multi_client.go
```diff
@@ -87,77 +89,103 @@ func NewMultiRPCClientFromEndpoints(
...
+			} else { // the batch call succeeded, and the eth_blockNumber call had results.\
+				// try to get the block number.
+				r, ok := req[blockNumReqIndex].Result.(*string)
+				if !ok {
+					errs[i] = fmt.Errorf("result from eth_blockNumber was not a string")
+					m.logger.Debug(
+						"result from eth_blockNumber was not a string",
+						zap.String("url", url),
+					)
+				} else {
```
technicallyty, line 144: this is kind of a nasty if/else chain.. which could be solved by using errgroup, however errgroup only gives you the first non-nil error, instead of all of them, which the code before did. so opting to keep that the same here.

but if having _all_ the errors is not that important, i can clean this up by using errgroup. that i way i can just log and then return err when needed, reducing a LOT of if/else junk.

providers/apis/defi/ethmulticlient/multi_client.go
```diff
@@ -87,77 +89,104 @@ func NewMultiRPCClientFromEndpoints(
...
+			results[i] = result{height, req[:blockNumReqIndex]}
+		}(clientIdx)
+	}
+	wg.Wait()
 
-			maxHeight = newHeight
-			copy(batchElems, req[:blockNumReqIndex])
+	// see which of the results had the largest height, and store the index of that result.
+	var maxHeight uint64
+	var maxHeightIndex int
```
aljo242:
```suggestion
	var (
	    maxHeight uint64
	    maxHeightIndex int
	)
```

providers/apis/defi/ethmulticlient/multi_client.go
```diff
@@ -87,77 +89,104 @@ func NewMultiRPCClientFromEndpoints(
 
 // BatchCallContext injects a call to eth_blockNumber, and makes batch calls to the underlying EVMClients.
 // It returns the first response it sees from a node which has the greatest height. An error is returned
```
aljo242: Is this line true?

Don't don't we always look thru all responses and then choose the highest one?

technicallyty: good catch - correct, we analyze all responses before returning the highest

technicallyty: fixed here [fix comment](https://github.com/skip-mev/slinky/pull/495/commits/d1f8c737579fc5d415cc3e61e2b53bf173fa2443)

Eric-Warehime (approved):

Eric-Warehime:
We don't have great coverage around running mutliple eth clients against uniswap markets in k8s right now. Would be nice to test everything locally for now.

aljo242 (approved):

technicallyty:
> We don't have great coverage around running mutliple eth clients against uniswap markets in k8s right now. Would be nice to test everything locally for now.

tested locally with a few base providers. watched logs of go routines entering and exiting. going to merge this now 🏎️

PR #498, opened 2024-06-06
fix(ci): build not only on tag push, but create too

Zygimantass:


Zygimantass:
Through some experimentation I discovered that there's a difference between the create and push events. What we do is most often just creating a tag on HEAD of main, but push needs a commit that is pushed on the tag afterwards. As such, to be safe, I covered both bases - you probably will not need to run the build workflows manually on release.

.github/workflows/build.yml
```diff
@@ -1,5 +1,8 @@
 name: Build Slinky docker images
 on:
+  create:
+    tags:
+      - "v.*"
```
davidterpay: Should this be `v*`?

Zygimantass: i think youre right, i dont think we even need this create thing lol

aljo242 (approved):

davidterpay (approved):

PR #503, opened 2024-06-10
refactor: consolidate oracle and provider orchestrator

technicallyty:
- folds the logic from orchestrator into Oracle. this removes some of the hairy indirection between the two services. 
- OracleServer no longer starts the oracle. the `main` function starts the oracle, and then passes it into the server


pkg/math/testutils/median.go
```diff
@@ -38,6 +39,8 @@ func (m *MedianAggregator) SetProviderPrices(provider string, data types.Prices)
 	m.providerPrices[provider] = data
 }
 
+func (m *MedianAggregator) UpdateMarketMap(_ mmtypes.MarketMap) {}
```
technicallyty, line 42: the oracle used a general interface of an aggregator called `PriceAggregator`. in the orchestrator, it was explicitly the `IndexAggregator`. 

assuming we want to continue with the generalizable interface, i'm adding this method here. i'm not sure if it needs to do anything though.

service/servers/oracle/server_test.go
```diff
@@ -220,22 +217,3 @@ func (s *ServerTestSuite) TestOracleServerClose() {
...
-	// start the server, and expect immediate closure
-	go srv.StartServer(context.Background(), localhost, port)
-
-	// wait for server to close
-	select {
-	case <-srv.Done():
-	case <-time.After(1 * time.Second):
-		t.Fatal("server failed to stop")
-	}
-}
```
technicallyty, line 241: i have no idea what this was supposed to do so im just removing it 🤔

pkg/math/testutils/median.go
davidterpay: Yea this doesn't need to do anything. We don't use this aggregator and is mostly here just for testing purposes.

oracle/helpers_test.go
```diff
@@ -196,14 +196,33 @@ var (
...
+
+func (n noOpPriceAggregator) AggregatePrices() {
+}
+
+func (n noOpPriceAggregator) GetPrices() oracletypes.Prices {
+	return oracletypes.Prices{}
+}
+
+func (n noOpPriceAggregator) Reset() {
+}
```
davidterpay, line 217: Might be useful to move this into the testutils directory.

oracle/oracle.go
```diff
@@ -24,252 +27,131 @@ type Oracle interface {
 	IsRunning() bool
 	GetLastSyncTime() time.Time
 	GetPrices() types.Prices
-	GetMarketMap() *marketmaptypes.MarketMap
+	GetMarketMap() mmtypes.MarketMap
```
davidterpay: nit: can you move this and all other interfaces to the `interfaces.go` file.

oracle/oracle.go
```diff
@@ -24,252 +27,131 @@ type Oracle interface {
...
-		maxCacheAge:    time.Minute, // default max cache age is 1 minute
+// New returns a new Oracle.
+func New(
+	cfg config.OracleConfig,
+	aggregator PriceAggregator,
+	opts ...Option,
+) (Oracle, error) {
+	if err := cfg.ValidateBasic(); err != nil {
+		return nil, err
 	}
```
davidterpay, line 100: mega nit: would add nil check on the aggregator.

oracle/oracle.go
```diff
@@ -24,252 +27,131 @@ type Oracle interface {
...
+func (o *OracleImpl) GetMarketMapProvider() *mmclienttypes.MarketMapProvider {
+	return o.mmProvider
 }
 
-// GetLastSyncTime returns the last time the oracle successfully updated prices.
 func (o *OracleImpl) GetLastSyncTime() time.Time {
-	o.mtx.RLock()
-	defer o.mtx.RUnlock()
-
 	return o.lastPriceSync
```
davidterpay: Is there a reason we removed the read lock on this?

oracle/lifecycle.go
```diff
@@ -97,19 +118,18 @@ func (o *ProviderOrchestrator) execProviderFn(
...
 	o.mut.Lock()
 	defer o.mut.Unlock()
 
-	o.mainCtx, o.mainCancel = context.WithCancel(ctx)
 	return o.mainCtx, o.mainCancel
 }
 
-// getMainCtx returns the main context for the provider orchestrator.
-func (o *ProviderOrchestrator) getMainCtx() (context.Context, context.CancelFunc) {
+func (o *OracleImpl) setMainCtx(ctx context.Context) (context.Context, context.CancelFunc) {
```
davidterpay: nit: can we add back the godoc here.

oracle/market_mapper.go
```diff
@@ -10,9 +10,10 @@ import (
...
 
 // listenForMarketMapUpdates is a goroutine that listens for market map updates and
-// updates the orchestrated providers with the new market map.
-func (o *ProviderOrchestrator) listenForMarketMapUpdates(ctx context.Context) {
-	mmProvider := o.GetMarketMapProvider()
+// updates the orchestrated providers with the new market map. This method assumes a market map provider is present,
+// so callers of this method must nil check the provider first.
+func (o *OracleImpl) listenForMarketMapUpdates(ctx context.Context) {
+	mmProvider := o.mmProvider
 	ids := mmProvider.GetIDs()
```
davidterpay, line 17: We can probably just exit if `mmProvider` is nil.

oracle/options.go
```diff
@@ -1,91 +1,98 @@
...
+}
 
-		o.priceAggregator = agg
+func WithPriceProviders(pps ...*types.PriceProvider) Option {
+	return func(m *OracleImpl) {
+		for _, pp := range pps {
+			m.priceProviders[pp.Name()] = ProviderState{Provider: pp}
+		}
 	}
 }
```
davidterpay, line 88: Is this still necessary?

oracle/options.go
```diff
@@ -1,91 +1,98 @@
...
 
-// WithProviders sets the providers on the Oracle.
-func WithProviders(providers []*types.PriceProvider) Option {
-	return func(o *OracleImpl) {
-		o.providers = providers
+func WithMarketMapProvider(mmp *mmclienttypes.MarketMapProvider) Option {
+	return func(impl *OracleImpl) {
+		impl.mmProvider = mmp
 	}
```
davidterpay: nit: can you add go docs

cmd/slinky/main.go
```diff
@@ -254,50 +253,35 @@ func runOracle() error {
...
-		orchestrator.WithMarketMap(marketCfg),
-		orchestrator.WithPriceAPIQueryHandlerFactory(oraclefactory.APIQueryHandlerFactory),             // Replace with custom API query handler factory.
-		orchestrator.WithPriceWebSocketQueryHandlerFactory(oraclefactory.WebSocketQueryHandlerFactory), // Replace with custom websocket query handler factory.
-		orchestrator.WithMarketMapperFactory(oraclefactory.MarketMapProviderFactory),
-		orchestrator.WithAggregator(aggregator),
-	}
-	if updateMarketCfgPath != "" {
-		orchestratorOpts = append(orchestratorOpts, orchestrator.WithWriteTo(updateMarketCfgPath))
-	}
+	// Define the oracle options. These determine how the oracle are created & executed.
```
davidterpay:
```suggestion
	// Define the oracle options. These determine how the oracle is created & executed.
```

oracle/oracle.go
davidterpay: I think we still need this.

davidterpay:
lgtm, mostly nits. will wait for deployment metrics before we approval.

oracle/options.go
aljo242: Don't all oracle instances need to have a marketmap provider?



I think this should be a required argument rather than an `Option`

oracle/market_mapper.go
technicallyty: this function is only called if `o.mmProvider != nil`. i could still add it though for extra safety in case it gets called elsewhere in the future, but the original code did not check for this.

oracle/options.go
technicallyty: so this was to allow mock providers to be passed into the oracle. i don't think we can stub in fake provider responses without this. see `providers_test.go`.

oracle/options.go
technicallyty: actually, i forgot to remove this option. marketmap providers are instantiated via OracleConfig.Providers, so this option is unused now.

oracle/oracle.go
technicallyty: done [move interfaces to interface.go](https://github.com/skip-mev/slinky/pull/503/commits/12ba0b0d4f7fca0621b4a5a447631e9f476c61b7)

oracle/oracle.go
technicallyty: oops. fixed here [add read lock to GetLastSyncTime and nil check aggregator](https://github.com/skip-mev/slinky/pull/503/commits/8cb65abfcb2f0d9f70bfcc91acb30b11a7416198)

oracle/oracle.go
technicallyty: kek i dont know why this got removed, added back here [add read lock to GetLastSyncTime and nil check aggregator](https://github.com/skip-mev/slinky/pull/503/commits/8cb65abfcb2f0d9f70bfcc91acb30b11a7416198)

oracle/lifecycle.go
technicallyty: https://github.com/skip-mev/slinky/pull/503/commits/bf436df5a1b3bf7fd24808bae54a765834ee8c44

oracle/options.go
technicallyty: added godocs to the other options [add godocs to other options](https://github.com/skip-mev/slinky/pull/503/commits/f9c7bd4f4371c2faa5e6a1bb2fc529736dcdded0)

oracle/helpers_test.go
technicallyty: i tend to prefer to keep stuff like this in the package where its used, and only move it out when some other package needs it. happy to move it out if you disagree though

aljo242 (approved):
lgtm

davidterpay (approved):

PR #512, opened 2024-06-14
fix: Val Exits

davidterpay:


abci/strategies/aggregator/price_applier.go
```diff
@@ -78,6 +78,12 @@ func (opa *oraclePriceApplier) ApplyPricesFromVoteExtensions(ctx sdk.Context, re
 		return nil, err
 	}
 
+	opa.logger.Info(
```
Eric-Warehime: Debug pls

abci/ve/utils.go
```diff
@@ -208,8 +213,6 @@ func ValidateVoteExtensions(
 		if !cmtPubKey.VerifySignature(extSignBytes, vote.ExtensionSignature) {
 			return fmt.Errorf("failed to verify validator %X vote extension signature", valConsAddr)
 		}
-
-		sumVP += vote.Validator.Power
```
Eric-Warehime, line 212: ~~We need this.~~

davidterpay: kek

abci/ve/utils.go
```diff
@@ -182,10 +182,15 @@ func ValidateVoteExtensions(
...
+		// If the validator does not have a valid public key, we skip the signature verification logic but still include
+		// the validator's voting power in the total voting power. The app may have pruned the validator's public key
+		// from the store, but comet considered the validator as active and included them in the commit since there
+		// is a 1 block delay between the validator set update on the app and comet.
+		sumVP += vote.Validator.Power
 		valConsAddr := sdk.ConsAddress(vote.Validator.Address)
 		pubKeyProto, err := valStore.GetPubKeyByConsAddr(ctx, valConsAddr)
 		if err != nil {
-			return fmt.Errorf("failed to get validator %X public key: %w", valConsAddr, err)
+			continue
```
g3ndo9kar9, line 193: can we check that this is a DNE error before continuing?

tests/integration/slinky_validator_suite.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		oracle := GetOracleSideCar(node)
+		SetOracleConfigsOnOracle(oracle, oracleConfig)
+		s.Require().NoError(RestartOracle(node))
+	}
+
+	// wait for the oracle to start
+	s.T().Log("Waiting for oracle to start...")
+	time.Sleep(45 * time.Second)
```
g3ndo9kar9, line 84: why do we have a sleep here? Is there no way to block on whether a price is included in these VEs?

Eric-Warehime (approved):

abci/ve/utils.go
technicallyty, line 193: should we have a comment on why we made this change?

davidterpay: I think the comment above should cover it. do u think there is anything else i should add to the comment?

technicallyty: oh whoops, i missed that lol 😪😪 😪😪

g3ndo9kar9 (approved):
lgtm

PR #534, opened 2024-06-19
feat: currency pair mapping query for x/oracle

technicallyty:
adds a grpc query to x/oracle that returns mapping of `currency_pair_id (uint64)` -> `curency_pair`. 

this is needed for indexers to make sense of the `currency_pair_id` that is available in the vote extension data when querying tendermint blocks.

PS i also ran `make proto-all`. diff is large cuz of that


x/oracle/keeper/keeper.go
```diff
@@ -286,6 +286,17 @@ func (k *Keeper) GetAllCurrencyPairs(ctx sdk.Context) []slinkytypes.CurrencyPair
 	return cps
 }
 
+// GetCurrencyPairMapping returns a CurrencyPair mapping by ID that have currently been stored to state.
+func (k *Keeper) GetCurrencyPairMapping(ctx sdk.Context) map[uint64]slinkytypes.CurrencyPair {
+	pairs := make(map[uint64]slinkytypes.CurrencyPair)
```
Eric-Warehime:
```suggestion
	pairs := make(map[uint64]slinkytypes.CurrencyPair, k.numCPs)
```

Eric-Warehime (approved):

aljo242 (approved):

PR #540, opened 2024-06-24
feat: Add delete market to keeper, for future proofing

Eric-Warehime:
Add delete method to x/mm keeper, so that in the future we can use this to remove markets in upgrade handlers. 

I haven't added any actual handling/msgs since we currently don't want to do this via  on chain txn.

closes BLO-1387

### DIFF
--- a/x/marketmap/keeper/keeper.go
+++ b/x/marketmap/keeper/keeper.go
@@ -118,7 +118,9 @@ func (k *Keeper) UpdateMarket(ctx sdk.Context, market types.Market) error {
 	return k.markets.Set(ctx, types.TickerString(market.Ticker.String()), market)
 }
 
-// DeleteMarket updates a Market.
+// DeleteMarket removes a Market.
+// This is currently only expected to be called in upgrade handlers, and callers will need to separately call
+// RemoveCurrencyPair on x/oracle to clean up leftover state in that module.
 func (k *Keeper) DeleteMarket(ctx sdk.Context, tickerStr string) error {
 	// Check if Ticker exists
 	alreadyExists, err := k.markets.Has(ctx, types.TickerString(tickerStr))
--- a/x/marketmap/keeper/keeper.go
+++ b/x/marketmap/keeper/keeper.go
@@ -128,7 +128,6 @@ func (k *Keeper) DeleteMarket(ctx sdk.Context, tickerStr string) error {
 	if !alreadyExists {
 		return types.NewMarketDoesNotExistsError(types.TickerString(tickerStr))
 	}
-	// Create the config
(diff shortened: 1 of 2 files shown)


x/marketmap/keeper/keeper.go
```diff
@@ -118,6 +118,20 @@ func (k *Keeper) UpdateMarket(ctx sdk.Context, market types.Market) error {
...
+func (k *Keeper) DeleteMarket(ctx sdk.Context, tickerStr string) error {
+	// Check if Ticker exists
+	alreadyExists, err := k.markets.Has(ctx, types.TickerString(tickerStr))
+	if err != nil {
+		return err
+	}
+	if !alreadyExists {
+		return types.NewMarketDoesNotExistsError(types.TickerString(tickerStr))
+	}
+	// Create the config
```
technicallyty:
```suggestion
```

technicallyty (commented):
is there a possible error scenario here?

- market gets deleted
- somewhere down the line an authority wants to add a market back, which calls the `AfterMarketCreated` hook
- `AfterMarketCreated` fails saying CurrencyPair already exists, even though it was deleted from MM?

do we also need a AfterMarketDeleted hook?

aljo242 (approved):

Eric-Warehime:
> is there a possible error scenario here?
> 
> * market gets deleted
> * somewhere down the line an authority wants to add a market back, which calls the `AfterMarketCreated` hook
> * `AfterMarketCreated` fails saying CurrencyPair already exists, even though it was deleted from MM?
> 
> do we also need a AfterMarketDeleted hook?

I think we don't want to manage that in hooks. I expect that the DeleteMarket calls will all only happen in upgrade handlers for now, and I would also expect that any x/oracle state should be cleaned up at the same time via `RemoveCurrencyPair` calls.

cc @technicallyty lmk if you agree or not.

technicallyty:
> > is there a possible error scenario here?
> > 
> > * market gets deleted
> > * somewhere down the line an authority wants to add a market back, which calls the `AfterMarketCreated` hook
> > * `AfterMarketCreated` fails saying CurrencyPair already exists, even though it was deleted from MM?
> > 
> > do we also need a AfterMarketDeleted hook?
> 
> I think we don't want to manage that in hooks. I expect that the DeleteMarket calls will all only happen in upgrade handlers for now, and I would also expect that any x/oracle state should be cleaned up at the same time via `RemoveCurrencyPair` calls.
> 
> cc @technicallyty lmk if you agree or not.

sounds fine to me. maybe lets throw a comment on the method to warn callers to clean up x/oracle state?

technicallyty (approved):

PR #546, opened 2024-06-25
feat: Sidecar Development Compose

davidterpay:
Adding a separate compose that generates markets and will only spin up the sidecar along with grafana and prometheus. Expecting this to be used for people who only want to use the sidecar.


g3ndo9kar9 (approved):
lgtm

aljo242 (approved):

PR #548, opened 2024-06-25
feat: add upsert markets API [BLO-1400]

g3ndo9kar9:
## In This PR
- I implement `MsgUpsertMarkets` as a separate API to `Create` / `Update` APIs in market-map
## Let's Discuss
- I preferred to introduce as a separate API to avoid any backwards incompat issue (i.e API / contract breaking changes)
- Considering not many other ppl are using these APIs in the wild currently, I'm also happy to either deprecate `Create / Update` or replace `Update` with `Upsert`


x/marketmap/keeper/msg_server.go
```diff
@@ -123,6 +173,32 @@ func (ms msgServer) UpdateMarkets(goCtx context.Context, msg *types.MsgUpdateMar
...
+// verifyMarketAuthoritiesAndExtractContext verifies that the msg-submitter is a market-authority
+// and returns the context for the msg, this method returns an error if the submitter is not a market
+// authority.
+func (ms msgServer) verifyMarketAuthoritiesAndExtractContext(goCtx context.Context, msg interface {
+	GetAuthority() string
+},
+) (sdk.Context, error) {
+	if msg == nil {
+		return sdk.Context{}, fmt.Errorf("unable to process nil msg")
+	}
```
technicallyty: technically you don't even need this, as gRPC code will always call your handlers with a default value request struct

example from generated pb code:

```go
func _Msg_AddCurrencyPairs_Handler(srv interface{}, ctx context.Context, dec func(interface{}) error, interceptor grpc.UnaryServerInterceptor) (interface{}, error) {
	in := new(MsgAddCurrencyPairs) // <--- default value message constructed using `new` here
	if err := dec(in); err != nil {
		return nil, err
	}
	if interceptor == nil {
		return srv.(MsgServer).AddCurrencyPairs(ctx, in)
	}
	info := &grpc.UnaryServerInfo{
		Server:     srv,
		FullMethod: "/slinky.oracle.v1.Msg/AddCurrencyPairs",
	}
	handler := func(ctx context.Context, req interface{}) (interface{}, error) {
		return srv.(MsgServer).AddCurrencyPairs(ctx, req.(*MsgAddCurrencyPairs))
	}
	return interceptor(ctx, in, info, handler)
}
```

x/marketmap/keeper/msg_server.go
```diff
@@ -23,23 +23,83 @@ func NewMsgServer(k *Keeper) types.MsgServer {
...
+		ctx.EventManager().EmitEvent(event)
 	}
 
-	found := checkMarketAuthority(msg.Authority, params)
-	if !found {
-		return nil, fmt.Errorf("request signer %s does not match module market authorities", msg.Authority)
+	// validate that the new state of the marketmap is valid
+	if err = ms.k.ValidateState(ctx, msg.Markets); err != nil {
+		return nil, err
+	}
```
technicallyty: seems like this would be better to have in the beginning of the function. no point in doing reads/writes if the markets in the message were bad

technicallyty (approved):

x/marketmap/keeper/msg_server.go
technicallyty: oh nvm looks like this is some sort of stateful check?

technicallyty:
should we add some `ValidateBasic` impls for the new message in`types/msg.go`?

x/marketmap/keeper/msg_server.go
g3ndo9kar9: ya, this is a verification that the current-state + applied updates is valid. Agreed tho, that we don't have a nice reversion if the ultimate validation fails (we have this when applying message servers, but it kinda sucks that we have dirty state every time we apply a message)

x/marketmap/keeper/msg_server.go
g3ndo9kar9: I saw the nil check on the other msg handlers, so I opted to replicate that check in `UpsertMarkets` + isolate into a helper fn. LMK if you'd prefer that I remove the nil check across all msg-handlers

g3ndo9kar9:
> should we add some ValidateBasic impls for the new message intypes/msg.go?

Updated

x/marketmap/keeper/msg_server.go
technicallyty: i see - i can just open a ticket to address this. probs has some cascading changes with tests and such

technicallyty: https://linear.app/skip/issue/BLO-1431/remove-unnecessary-nil-check-from-grpc-handlers

x/marketmap/keeper/msg_server.go
```diff
@@ -123,6 +173,32 @@ func (ms msgServer) UpdateMarkets(goCtx context.Context, msg *types.MsgUpdateMar
 	return &types.MsgUpdateMarketsResponse{}, ms.k.SetLastUpdated(ctx, uint64(ctx.BlockHeight()))
 }
 
+// verifyMarketAuthoritiesAndExtractContext verifies that the msg-submitter is a market-authority
+// and returns the context for the msg, this method returns an error if the submitter is not a market
+// authority.
+func (ms msgServer) verifyMarketAuthoritiesAndExtractContext(goCtx context.Context, msg interface {
```
aljo242: Why does this function need to return a context?

can't we just pass an sdk.Context into here always? rather than passing a context.Context

x/marketmap/types/msg.go
```diff
@@ -11,8 +11,30 @@ var (
...
+
+	if len(m.Markets) == 0 {
+		return fmt.Errorf("no markets to upsert")
+	}
+
+	for _, market := range m.Markets {
+		if err := market.ValidateBasic(); err != nil {
+			return err
+		}
+	}
```
aljo242: Could we check for duplicate markets here?

x/marketmap/keeper/msg_server.go
g3ndo9kar9: I wanted to reduce code duplication across the handlers, but I can unwrap context before if u prefer

x/marketmap/types/msg.go
g3ndo9kar9: We aren't doing this for the other msg-types (`Create` / `Update`), but I can add

aljo242 (approved):

PR #557, opened 2024-06-27
feat: dYdX CMC Research

davidterpay:


providers/apis/dydx/multi_market_map_fetcher.go
```diff
@@ -1,186 +0,0 @@
-package dydx
```
davidterpay, line 1: Redundant as the research json already has mainnet markets. Moving forward deployments are going to be:

1. dYdX staging (w/o VPN) pointing to mainnet
2. dYdX staging (w VPN) pointing to mainnet
3. dYdX research (w VPN)
4. dYdX research (w/o VPN)
5. dYdX research coinmarketcap

providers/apis/dydx/parse.go
```diff
@@ -49,6 +50,7 @@ var ProviderMapping = map[string]string{
 	"Raydium":              raydium.Name,
 	"UniswapV3-Ethereum":   uniswapv3.ProviderNames[constants.ETHEREUM],
 	"UniswapV3-Base":       uniswapv3.ProviderNames[constants.BASE],
+	coinmarketcap.Name:     coinmarketcap.Name,
```
davidterpay, line 53: Needs to be added here since the existing logic inherits this class. Once we migrate to v6 we should move this logic directly into the research provider.

aljo242 (approved):

providers/apis/dydx/research_api_handler.go
```diff
@@ -136,12 +140,16 @@ func (h *ResearchAPIHandler) ParseResponse(
 
 // researchJSONToQueryAllMarketsParamsResponse converts a dydx research json into a
 // QueryAllMarketsParamsResponse.
-func researchJSONToQueryAllMarketsParamsResponse(research dydxtypes.ResearchJSON) (dydxtypes.QueryAllMarketParamsResponse, error) {
+func (h *ResearchAPIHandler) researchJSONToQueryAllMarketsParamsResponse(research dydxtypes.ResearchJSON) (dydxtypes.QueryAllMarketParamsResponse, error) {
 	// iterate over all entries in the research json + unmarshal it's market-params
 	resp := dydxtypes.QueryAllMarketParamsResponse{}
 	for _, market := range research {
+		if market.CMCID < 0 {
```
g3ndo9kar9: Does this limit us (even the normal research-json provider) to ignore any market that does not have a cmc-id?

providers/apis/dydx/multi_market_map_fetcher.go
```diff
@@ -142,29 +158,71 @@ func (f *MultiMarketMapRestAPIFetcher) Fetch(ctx context.Context, chains []mmcli
...
 			}
 		}
-	} else {
-		return dydxResearchMarketMapResponse
+	}
+
+	// if the fetcher is only for CoinMarketCap markets, filter out all non-CMC markets
+	if f.isCMCOnly {
+		for ticker, market := range dydxMainnetMarketMap.Markets {
+			market.Ticker.MinProviderCount = 1
```
g3ndo9kar9, line 192: why do we need to do this? Is this meant to over-ride all markets that exist on mainnet, but not in research-json?

providers/apis/dydx/multi_market_map_fetcher.go
```diff
@@ -142,29 +158,71 @@ func (f *MultiMarketMapRestAPIFetcher) Fetch(ctx context.Context, chains []mmcli
...
+			dydxMainnetMarketMap.Markets[ticker] = market
+
+			var (
+				seenCMC     = false
+				cmcProvider mmtypes.ProviderConfig
+			)
+
+			for _, provider := range market.ProviderConfigs {
```
g3ndo9kar9, line 200: When would a mainnet market ever have the CMC provider?

providers/apis/dydx/multi_market_map_fetcher.go
```diff
@@ -142,29 +158,71 @@ func (f *MultiMarketMapRestAPIFetcher) Fetch(ctx context.Context, chains []mmcli
...
+				market.ProviderConfigs = []mmtypes.ProviderConfig{cmcProvider}
+				dydxMainnetMarketMap.Markets[ticker] = market
+				continue
+			}
+
+			// If we did not see a CMC provider, we can attempt to add it using the CMC marketmap
+			cmcMarket, ok := marketmaps.CoinMarketCapMarketMap.Markets[ticker]
+			if !ok {
+				f.logger.Info("did not find CMC market for ticker", zap.String("ticker", ticker))
+				delete(dydxMainnetMarketMap.Markets, ticker)
```
g3ndo9kar9, line 218: Does this mean we can create invalid market-maps? I.e we remove an adjust by

g3ndo9kar9 (commented):
comments

providers/apis/dydx/multi_market_map_fetcher.go
davidterpay: this will only over-ride if its meant to be a CMC provider only. so research json would still have the correct min provider count.

providers/apis/dydx/multi_market_map_fetcher.go
davidterpay: it wouldn't which is why we just cross reference with what is in our constants. the markets being added rn via the frontend should have CMC IDs.

providers/apis/dydx/multi_market_map_fetcher.go
davidterpay: technically yes, but all of the important adjust bys are in the constants dir. this is a temp fix till we fully move to marketmap generation.

providers/apis/dydx/research_api_handler.go
davidterpay: good catch, should also check that its cmc

PR #566, opened 2024-07-01
feat: Switch over provider

davidterpay:
This provider allows dYdX validators to supply their x/prices and x/marketmap endpoints such that we can gracefully switch over to `x/marketmap` once it is live on mainnet/testnet.


providers/apis/dydx/switch_over_fetcher.go
```diff
@@ -0,0 +1,153 @@
...
+	pricesFetcher mmclient.MarketMapFetcher
+	// marketmapFetcher is the fetcher for the x/marketmap market map.
+	marketmapFetcher mmclient.MarketMapFetcher
+	// switched is true if the fetcher has switched over to the x/marketmap API.
+	switched bool
+}
+
+// DefaultSwitchOverMarketMapFetcher returns a new SwitchOverProvider with the default
+// dYdX x/prices and x/marketmap fetchers.
+func DefaultSwitchOverMarketMapFetcher(
```
technicallyty: `NewDefaultSwitchOverMarketMapFetcher`?

providers/apis/dydx/switch_over_fetcher.go
```diff
@@ -0,0 +1,153 @@
...
+) (mmclient.MarketMapFetcher, error) {
+	if logger == nil {
+		return nil, fmt.Errorf("logger is nil")
+	}
+	if api.Name != SwitchOverAPIHandlerName {
+		return nil, fmt.Errorf("expected api name %s, got %s", SwitchOverAPIHandlerName, api.Name)
+	}
+	if len(api.Endpoints) != 2 {
+		return nil, fmt.Errorf(
+			"expected two endpoints, got %d; the first end point must be the rest port second endpoint gRPC",
```
technicallyty:
```suggestion
			"expected two endpoints, got %d; the first endpoint must be the rest port second endpoint gRPC",
```
is it correct the first endpoint must be a rest port? little confused on what they're supposed to be by this error

providers/apis/marketmap/client.go
```diff
@@ -62,6 +62,19 @@ func NewGRPCClient(
...
+	conn *grpc.ClientConn,
+	api config.APIConfig,
+	apiMetrics metrics.APIMetrics,
+) (mmtypes.QueryClient, error) {
+	return &MarketMapClient{
+		QueryClient: mmtypes.NewQueryClient(conn),
+		apiMetrics:  apiMetrics,
+		api:         api,
+	}, nil
+}
```
technicallyty: theres some validation stuff done on `ApiConfig` and `ApiMetrics` in `NewGRPCClient` above ; do we need to do that here as well?

aljo242 (approved):

providers/apis/dydx/switch_over_fetcher.go
```diff
@@ -0,0 +1,153 @@
...
+) mmclient.MarketMapResponse {
+	if f.switched {
+		return f.marketmapFetcher.Fetch(ctx, chains)
+	}
+
+	f.logger.Debug("fetching marketmap from x/marketmap")
+	resp := f.marketmapFetcher.Fetch(ctx, chains)
+	if len(resp.Resolved) > 0 {
+		f.logger.Info("got response from x/marketmap; switching over to x/marketmap")
+		f.switched = true
```
g3ndo9kar9, line 174: We'll probs want to know when nodes have switched over to the x/prices market-map. Will there be an easy way to do this?

providers/apis/dydx/utils.go
```diff
@@ -58,6 +61,25 @@ var DefaultAPIConfig = config.APIConfig{
...
+	Timeout:          20 * time.Second, // Set a high timeout to account for slow API responses in the case where many markets are queried.
+	Interval:         10 * time.Second,
+	ReconnectTimeout: 2000 * time.Millisecond,
+	MaxQueries:       1,
+	Endpoints: []config.Endpoint{
+		{
+			URL: "https://dydx-api.lavenderfive.com", // REST endpoint (HTTP/HTTPS prefix)
+		},
+		{
+			URL: "localhost:9090", // gRPC endpoint (NO HTTP/HTTPS prefix)
```
g3ndo9kar9, line 78: If we're using a 3rd party REST endpoint why are we using a local address for GRPC? I'd suggest that we use local addresses for both

providers/base/api/handlers/request_handler.go
```diff
@@ -30,6 +33,9 @@ type RequestHandlerImpl struct {
 	headers map[string]string
 }
 
+// NoOpRequestHandler is a no-op RequestHandler that does nothing.
+type NoOpRequestHandler struct{}
```
g3ndo9kar9: Where are we using this?

g3ndo9kar9 (approved):
lgtm w/ comments / qs

providers/apis/dydx/switch_over_fetcher.go
davidterpay: yea supposed to be rest and gRPC

davidterpay: ill just remove that from here for now and put it in docs instead

providers/apis/dydx/switch_over_fetcher.go
davidterpay: yes, the marketmapclient has metrics associated with it, specifically RPC metrics. So we should be able to see when the switch over happens.

providers/apis/dydx/utils.go
davidterpay: had as a placeholder while i was testing. ill just default to 1317 and 9090

providers/base/api/handlers/request_handler.go
davidterpay: testing

g3ndo9kar9: Why can't we use mocks instead of a Nop implementation

PR #567, opened 2024-07-01
feat: WrappedPreBlocker

technicallyty:
previously, simapp would not work with x/upgrade handlers. this is because upgrades rely on `PreBlocker` but this was being overwritten with the oracle PreBlocker.

fixed by creating a new method on oracle's PreBlockHandler: "WrappedPreBlocker" which takes a `module.Manager`, and call the PreBlock function to trigger upgrades. h/t @aljo242 for showing me the way

i added a comment to the existing PreBlocker that notes it is deprecated and to use WrappedPreBlocker instead.


tests/simapp/app.go
```diff
@@ -530,6 +531,28 @@ func (app *SimApp) RegisterAPIRoutes(apiSvr *api.Server, apiConfig config.APICon
 	}
 }
 
+// wrapAndSetPreBlocker sets the application's PreBlocker while still calling the originals.
+// This is needed to set custom PreBlock logic, but still allow things like x/upgrade to function.
+func (app *SimApp) wrapAndSetPreBlocker(pb sdk.PreBlocker) {
```
g3ndo9kar9: Let's export this fn publicly? This is an operation all integrators have to roll on their own currently

g3ndo9kar9 (approved):
lgtm

tests/simapp/app.go
technicallyty: i could make it so `oraclePreBlockHandler.PreBlocker()` takes a `*module.Manager` argument, so this does it automatically? does that sound good?

aljo242: Yeah, that makes sense to me

technicallyty: updated PR to reflect the change, let me know if all is good @nivasan1 @aljo242

davidterpay (approved):

abci/preblock/oracle/preblock_test.go
technicallyty, line 1: i just switched all the tests over to the non-deprecated version of the method, and added a case to ensure `PreBlock` from `ModuleManager` is actually called. lmk if this is ideal or not!

aljo242 (approved):

PR #572, opened 2024-07-02
fix: okx 

aljo242:
Combine `CreateMessages` calls into batches for OKX


providers/websockets/okx/messages.go
```diff
@@ -24,21 +24,18 @@ type (
...
 	// EventSubscribe is the event denoting that we have successfully subscribed to a channel.
 	EventSubscribe EventType = "subscribe"
 	// EventTickers is the event for tickers. By default, this field will not be populated
 	// in a properly formatted message. So we set the default value to an empty string.
 	EventTickers EventType = ""
 	// EventError is the event for an error.
 	EventError EventType = "error"
+
+	// subscriptionMsgSize is the amount of tickers to subscribe to in a single Subscribe message.
+	subscriptionMsgSize = 10
```
aljo242, line 38: arbitrary

g3ndo9kar9: Stuff like this belongs in the `WebsocketConfig` IMO, this way we can have all other providers use a default value (1), but okx moves to a larger # of tickers per sub

aljo242:
Closes as we have https://github.com/skip-mev/slinky/pull/589

PR #584, opened 2024-07-09
feat: market map enable helper functions

aljo242:
Closes: BLO-1505


technicallyty (approved):
LGTM. they only needed this in keeper? they didnt want it exposed over grpc?

aljo242:
> LGTM. they only needed this in keeper? they didnt want it exposed over grpc?

yeah, they just want to have helper functions they can call from the keeper

PR #588, opened 2024-07-10
fix: add protection around marketmap hooks

aljo242:
- add nil protection check in Genesis
- init the keeper to have no-op hooks by default


technicallyty (approved):

davidterpay (approved):

davidterpay:
failing CI but otherwise lgtm

PR #589, opened 2024-07-10
feat: [BLO-1508] Configuring amount of subscriptions sent in a single batch/message

davidterpay:


providers/websockets/bybit/parse.go
```diff
@@ -26,6 +26,7 @@ func (h *WebSocketHandler) parseSubscriptionResponse(resp SubscriptionResponse)
 		return nil, nil
 	}
 
+	// TODO(david): Add a retry mechanism here.
```
aljo242, line 29: issue?

aljo242 (approved):

PR #597, opened 2024-07-12
fix: Subscribe to market data channels instead of public for OKX

davidterpay:


aljo242 (approved):

aljo242:
@davidterpay conflicts

PR #598, opened 2024-07-15
feat: Price Daemon

davidterpay:


oracle/config/app.go
```diff
@@ -65,6 +85,12 @@ type AppConfig struct {
 
 	// MetricsEnabled is a flag that determines whether oracle metrics are enabled.
 	MetricsEnabled bool `mapstructure:"metrics_enabled" toml:"metrics_enabled"`
+
+	// MaxAge is the maximum age of the latest price response before it is considered stale.
+	MaxAge time.Duration `mapstructure:"max_age" toml:"max_age"`
```
Eric-Warehime: `oracle.max_age` feels pretty vague in terms of describing what this does. How about `oracle.price_ttl` instead or something along those lines?

oracle/config/app.go
```diff
@@ -74,11 +100,23 @@ func (c *AppConfig) ValidateBasic() error {
...
+	if c.MaxAge <= 0 || c.MaxAge > MaxMaxAge {
+		return fmt.Errorf("poorly formatted app.toml (oracle subsection): oracle max age must be between 0 and %s", MaxMaxAge)
+	}
+
+	if c.Interval <= 0 || c.Interval > MaxInterval {
+		return fmt.Errorf("poorly formatted app.toml (oracle subsection): oracle interval must be between 0 and %s", MaxInterval)
+	}
+
+	if c.Interval >= c.MaxAge {
+		return fmt.Errorf("poorly formatted app.toml (oracle subsection): oracle interval must be strictly less than max age")
```
Eric-Warehime: I think it's good to have these validations, but we should be setting defaults for people. Nobody should really be touching these on their own.

Eric-Warehime (commented):
LGTM, but we should set default values and make this backwards compatible.

Eric-Warehime (approved):

.github/workflows/e2e.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.22.5
+          go-version: 1.22.3
```
aljo242: why this change?

oracle/config/app.go
```diff
@@ -120,27 +150,53 @@ func ReadConfigFromAppOpts(opts servertypes.AppOptions) (AppConfig, error) {
 		if cfg.Enabled, err = cast.ToBoolE(v); err != nil {
 			return cfg, err
 		}
+	} else {
+		cfg.Enabled = DefaultOracleEnabled // set to default
```
aljo242: can we have all of these values set to default before we overwrite w flags

service/clients/oracle/daemon.go
```diff
@@ -0,0 +1,164 @@
...
+			time.Since(ts).String(),
+		)
+	}
+
+	return latest, nil
+}
+
+// Stop stops the price daemon.
+func (d *PriceDaemon) Stop() error {
+	d.doneCh <- struct{}{}
```
aljo242: should close this channel too

service/clients/oracle/daemon.go
```diff
@@ -0,0 +1,164 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			d.logger.Info("stopping price daemon from context")
+			return ctx.Err()
+		case <-d.doneCh:
+			d.logger.Info("price daemon stopped")
+			close(d.doneCh)
```
aljo242: I would close the channel in the `Stop()` function.  This sends a final signal on the channel so this case will still be met

.github/workflows/e2e.yml
davidterpay: thought this was related to to the docker issue, i can revert back to .5.

service/clients/oracle/daemon.go
davidterpay: the channel gets closed in `Stop()`. can close it here as well but issue is that this would block if start isn't executed first.

PR #603, opened 2024-07-16
feat: Parse out CMC markets from MMU and convert to CMC only market map

davidterpay:


cmd/slinky/main.go
```diff
@@ -343,3 +357,56 @@ func overwriteMarketMapEndpoint(cfg config.OracleConfig, overwrite string) (conf
 
 	return cfg, fmt.Errorf("no market-map provider found in config")
 }
+
+// filterToOnlyCMCMarkets is a helper function that filters out all markets that are not from CoinMarketCap. It
+// mutates the marketmap to only include CoinMarketCap markets.
+func filterToOnlyCMCMarkets(marketmap mmtypes.MarketMap) mmtypes.MarketMap {
```
aljo242: just to note that this is for the dydx chain so all of these IDs are just the ID of the BASE asset. So they will always be prices in BASE/USD

cmd/slinky/main.go
```diff
@@ -343,3 +357,57 @@ func overwriteMarketMapEndpoint(cfg config.OracleConfig, overwrite string) (conf
...
+	return res
+}
+
+type metaDataJSON struct {
+	AggregateIDs []aggregateIDsJSON `json:"aggregate_ids"`
+}
+type aggregateIDsJSON struct {
+	Venue string `json:"venue"`
+	ID    string `json:"ID"`
+}
```
Eric-Warehime: Any reason we shouldn't be using https://github.com/skip-mev/slinky/commit/df1036cf5ead09a2e20f4165857fe14b4314b197 here?

cmd/slinky/main.go
```diff
@@ -176,6 +179,13 @@ func init() {
 		"",
 		"Use a custom listen-to endpoint for market-map (overwrites what is provided in oracle-config).",
 	)
+	rootCmd.Flags().BoolVarP(
+		&useCMCOnly,
+		"use-cmc-only",
+		"",
+		false,
+		"Use CoinMarketCap only for price data. **This should only be used for testing.**. Only works if you pair this with the --market-config-path flag.",
+	)
```
Eric-Warehime: I really think this is a bad idea and despite the usage string, someone will use this incorrectly. The best way to avoid misuse is to put this in the config imo.

davidterpay: we could generate a CMC MM for every MM that we generate or separately run a script that can do the parsing logic.

davidterpay: Im assuming former probably isn't ideal

davidterpay: just FYI, we have several other fields in here where validators could incorrectly run things, is there a reason why you think this one is any different?

davidterpay: We could also try to obfuscate with env variables.

cmd/slinky/main.go
davidterpay: Did not realize that was in the code base.

cmd/slinky/main.go
davidterpay: I'm moving the logic into a separate script.

Eric-Warehime (approved):

PR #607, opened 2024-07-17
feat: osmosis provider

aljo242:
Closes: BLO-1537

```
{
  "prices": {
    "OSMO/USD": "54165419",
    "STARS/USD": "11621825831747928",
    "USDT/USD": "1000200040"
  },
  "timestamp": "2024-07-17T15:03:28.675080795Z"
}
```


Makefile
```diff
@@ -283,6 +285,7 @@ gen-mocks:
 	@echo "--> generating mocks"
 	@go install github.com/vektra/mockery/v2
 	@go generate ./...
+	@cd ./providers/apis/defi/osmosis && go generate ./...
```
technicallyty: shouldn't `go generate ./...` cover this?

pkg/http/round_tripper_with_headers.go
```diff
@@ -58,3 +60,53 @@ func (r *RoundTripperWithHeaders) RoundTrip(req *http.Request) (*http.Response,
...
+		opt(req)
+	}
+
+	resp, err := c.internal.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("http get: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New(http.StatusText(resp.StatusCode))
```
technicallyty: `resp.Status` contains `[code]: [status_text]`. could just use that here

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,235 @@
...
+	SpotPrice(grpcCtx context.Context,
+		poolID uint64,
+		baseAsset,
+		quoteAsset string,
+	) (SpotPriceResponse, error)
+}
+
+// ClientImpl is an implementation of a client to Osmosis using a
+// poolmanager Query Client.
+type ClientImpl struct {
```
technicallyty, line 36: nit: seems like this and `MultiClientImpl` don't need to be exported types

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,235 @@
...
+
+	}
+
+	wg.Wait()
+
+	return filterSpotPriceResponses(resps)
+}
+
+// filterSpotPriceResponses currently just chooses a random response as there is no way to differentiate.
+// TODO differentiate.
```
technicallyty: still todo?

providers/apis/defi/osmosis/price_fetcher.go
```diff
@@ -0,0 +1,198 @@
...
+			if err != nil {
+				pf.logger.Error("failed parse spot price response", zap.Error(err))
+				unresolved[ticker] = providertypes.UnresolvedResult{
+					ErrorWithCode: providertypes.NewErrorWithCode(
+						err,
+						providertypes.ErrorFailedToParsePrice,
+					),
+				}
+				return
+			}
```
technicallyty: shouldn't this have the unresolvedMtx lock?

maybe we should replace this logic with a closure:

```go
writeToUnresolved := func(ticker, err, code) {
    ....
}
```

providers/apis/defi/osmosis/client.go
aljo242: fixed

providers/apis/defi/osmosis/price_fetcher.go
technicallyty: yo @aljo242 any thoughts on the closure idea above?

providers/apis/defi/osmosis/client.go
technicallyty: comment still says TODO @aljo242

providers/apis/defi/osmosis/client.go
technicallyty: this was marked as resolved, but there wasn't a change @aljo242 , i noticed the other files have their client impl as exported types as well. so maybe fine for now, but could be a ticket for the future

providers/apis/defi/osmosis/price_fetcher.go
aljo242: pushed here https://github.com/skip-mev/slinky/pull/607/commits/bb7b13dd2d1d9a118c645a18ae70d789dbc6e577

pkg/http/round_tripper_with_headers.go
```diff
@@ -58,3 +60,53 @@ func (r *RoundTripperWithHeaders) RoundTrip(req *http.Request) (*http.Response,
...
+	if err != nil {
+		return nil, fmt.Errorf("http get: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New(resp.Status)
+	}
+
+	return resp, nil
+}
```
davidterpay, line 112: Why do we need this additional client? We already have a separate client in base/api/handlers/requesthandler?

davidterpay: can this functionality just be added there?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/skip-mev/slinky
 
-go 1.22
+go 1.22.3
```
davidterpay: can we make go verison bumps in separate PRs?

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,237 @@
...
+var (
+	_ Client = &ClientImpl{}
+	_ Client = &MultiClientImpl{}
+)
+
+// Client is the expected interface for an osmosis client.
+//
+//go:generate mockery --name Client --output ./mocks/ --case underscore
+type Client interface {
+	SpotPrice(grpcCtx context.Context,
```
davidterpay:
```suggestion
	SpotPrice(ctx context.Context,
```

providers/apis/defi/osmosis/types.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return tickerMetadata, nil
+}
+
+// DefaultAPIConfig is the default configuration for the Osmosis API price fetcher.
+var DefaultAPIConfig = config.APIConfig{
+	Enabled:          true,
+	Name:             Name,
+	Timeout:          4 * time.Second,
+	Interval:         1000 * time.Millisecond,
```
davidterpay: interval can probably be higher than 1 sec since osmo block times are higher.

providers/apis/defi/osmosis/types.go
```diff
@@ -0,0 +1,134 @@
...
+	ReconnectTimeout: 4 * time.Second,
+	MaxQueries:       10,
```
davidterpay, line 122: why do we need to allow 10 concurrent queries?

providers/apis/defi/osmosis/types.go
```diff
@@ -0,0 +1,134 @@
...
+	Atomic:           false,
+	BatchSize:        25, // maximal # of accounts in getMultipleAccounts query is 100
```
davidterpay: what is the batchsize doing?

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return SpotPriceResponse{}, err
+	}
+
+	var spotPriceResponse SpotPriceResponse
+	if err := json.NewDecoder(resp.Body).Decode(&spotPriceResponse); err != nil {
+		return SpotPriceResponse{}, err
+	}
+
+	c.apiMetrics.AddRPCStatusCode(c.api.Name, c.redactedURL, metrics.RPCCodeOK)
```
davidterpay: why are we recording RPC status codes when an http client is used?

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,237 @@
...
+			// Observe the latency of the request.
+			start := time.Now()
+			defer func() {
+				wg.Done()
+				mc.apiMetrics.ObserveProviderResponseLatency(mc.api.Name, metrics.RedactedEndpointURL(index), time.Since(start))
+			}()
+
+			resp, err := client.SpotPrice(ctx, poolID, baseAsset, quoteAsset)
+			if err != nil {
+				mc.apiMetrics.AddRPCStatusCode(mc.api.Name, metrics.RedactedEndpointURL(index), metrics.RPCCodeError)
```
davidterpay: same as above

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,237 @@
...
+				mc.logger.Error("failed to spot price in sub client", zap.String("url", url), zap.Error(err))
+				return
+			}
+
+			mc.apiMetrics.AddRPCStatusCode(mc.api.Name, metrics.RedactedEndpointURL(index), metrics.RPCCodeOK)
```
davidterpay: same as above

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,237 @@
...
+	perm := rand.Perm(len(responses))
+	for _, i := range perm {
+		resp := responses[perm[i]]
+		if resp.SpotPrice != "" {
+			return resp, nil
+		}
+	}
+
+	return SpotPriceResponse{}, fmt.Errorf("no responses found")
+}
```
davidterpay, line 230: if we are using multi-client doesn't it make more sense to also use query block height?

davidterpay: The eth and sol providers both do this.

providers/apis/defi/osmosis/price_fetcher.go
```diff
@@ -0,0 +1,204 @@
...
+	ctx context.Context,
+	tickers []oracletypes.ProviderTicker,
+) oracletypes.PriceResponse {
+	resolved := make(oracletypes.ResolvedPrices)
+	unresolved := make(oracletypes.UnResolvedPrices)
+
+	wg := sync.WaitGroup{}
+	unresolvedMtx := sync.Mutex{}
+	resolveMtx := sync.Mutex{}
+	wg.Add(len(tickers))
```
davidterpay: This will spin up an unbounded number of threads to make concurrent requests?

providers/apis/defi/osmosis/price_fetcher.go
```diff
@@ -0,0 +1,204 @@
...
+				))
+
+				return
+			}
+
+			resolvedTickerCallback(ticker, price)
+		}(ticker)
+	}
+
+	wg.Wait()
```
davidterpay: this should ideally be buffering a maximum number of concurrent requests otherwise we will certainly be rate limited.

pkg/http/round_tripper_with_headers.go
aljo242: this is an internal general http client and doesnt need all of the requestHandler functionality.  It makes sense to me to only have the Fetcher use a client that gives it the functionality that it needs

providers/apis/defi/osmosis/client.go
aljo242: This query does not return a block height so we would have to do an additional query alongside what we're doing

providers/apis/defi/osmosis/price_fetcher.go
aljo242: Now using an errgroup that limits the number of goroutines to `MaxQueries`

technicallyty (approved):
lgtm

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,237 @@
...
+	for i, client := range mc.clients {
+		url := mc.api.Endpoints[i].URL
+
+		index := i
+		go func(index int, client Client) {
+			// Observe the latency of the request.
+			start := time.Now()
+			defer func() {
+				wg.Done()
+				mc.apiMetrics.ObserveProviderResponseLatency(mc.api.Name, metrics.RedactedEndpointURL(index), time.Since(start))
```
davidterpay: is this metric going to be duplicated since the individual clients themselves already record this?

providers/apis/defi/osmosis/client.go
```diff
@@ -0,0 +1,237 @@
...
+		}(index, client)
+
+	}
+
+	wg.Wait()
+
+	return filterSpotPriceResponses(resps)
+}
+
+// filterSpotPriceResponses currently just chooses a random response as there is no way to differentiate.
```
davidterpay, line 215: should probably make a todo task out of this to see whether we want to also query block heights.

davidterpay (approved):
small nit

PR #609, opened 2024-07-17
chore: Remove testapp from install

Eric-Warehime:
Removes the testapp from the install directive. The install should just install the binary, and we have and use a separate make target for building and installing the test app binary.

BLO-1524


g3ndo9kar9 (approved):
lgtm

technicallyty (approved):

PR #611, opened 2024-07-18
fix: account for protocols that aren't tcp

g3ndo9kar9:
## In This PR
- I account for URL schemes that are not http (tcp, etc) when stripping schemes from GRPC target addresses

### DIFF
--- a/pkg/grpc/client.go
+++ b/pkg/grpc/client.go
@@ -1,7 +1,8 @@
 package grpc
 
 import (
-	"strings"
+	"fmt"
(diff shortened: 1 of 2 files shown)


technicallyty (approved):

aljo242 (commented):
Can we add a test to reproduce the issue that was occurring?

aljo242:
E2E tests are now failing due to this change

Eric-Warehime (approved):

PR #613, opened 2024-07-19
feat: Version RPC

davidterpay:


proto/slinky/service/v1/oracle.proto
```diff
@@ -41,4 +52,13 @@ message QueryMarketMapRequest {}
...
   // market_map defines the current market map configuration.
   slinky.marketmap.v1.MarketMap market_map = 1;
+}
+
+// QueryVersionRequest defines the request type for the Version method.
+message QueryVersionRequest {}
+
+// QueryVersionResponse defines the response type for the Version method.
+message QueryVersionResponse {
+  // version defines the current version of the oracle service.
```
aljo242:
```suggestion
  // Version defines the current version of the oracle service.
```

Comments on the proto should start capitalized like they will be in Go so that they can be proper godoc

aljo242 (approved):
lgtm with nit

PR #617, opened 2024-07-22
feat: `InitializeForGenesis`

aljo242:
Closes: BLO-1562

function interface implementation for dydx


technicallyty (approved):
game changing

PR #622, opened 2024-07-23
feat: Refresh prices on heartbeats

davidterpay:


oracle/types/provider.go
```diff
@@ -62,7 +66,9 @@ func (t DefaultProviderTicker) String() string {
 func NewProviderTickers(tickers ...ProviderTicker) ProviderTickers {
 	cache := make(map[string]ProviderTicker)
 	for _, ticker := range tickers {
+		cache[strings.ToLower(ticker.GetOffChainTicker())] = ticker
 		cache[ticker.GetOffChainTicker()] = ticker
+		cache[strings.ToUpper(ticker.GetOffChainTicker())] = ticker
```
aljo242, line 71: why do we need 3 entries in the cache?

aljo242 (approved):

PR #624, opened 2024-07-24
fix: App Config (app.toml) Parsing Fix

davidterpay:
app.toml automatically fills in fields with defaults even if they are not set so our parsing logic has to take the defaults into account.


aljo242 (approved):

oracle/config/app.go
```diff
@@ -152,17 +152,33 @@ func ReadConfigFromAppOpts(opts servertypes.AppOptions) (AppConfig, error) {
...
+		return cfg, nil
+	}
+
 	// get the oracle address
 	if v := opts.Get(flagOracleAddress); v != nil {
-		if cfg.OracleAddress, err = cast.ToStringE(v); err != nil {
-			return cfg, err
+		address, err := cast.ToStringE(v)
+		if err != nil {
+			return cfg, fmt.Errorf("oracle address must be a non-empty string")
```
technicallyty, line 163: this won't return an error in the empty case

oracle/config/app.go
```diff
@@ -175,15 +191,27 @@ func ReadConfigFromAppOpts(opts servertypes.AppOptions) (AppConfig, error) {
 
 	// get the price ttl
 	if v := opts.Get(flagPriceTTL); v != nil {
-		if cfg.PriceTTL, err = cast.ToDurationE(v); err != nil {
-			return cfg, err
+		priceTTL, err := cast.ToDurationE(v)
+		if err != nil {
+			return cfg, fmt.Errorf("price ttl must be a positive duration")
+		}
```
technicallyty, line 197: this won't return an error in negative durations, i.e. `cast.ToDuration("-5s")`

oracle/config/app.go
davidterpay: yea, we only replace if its non-empty. if its empty we just use the default.

oracle/config/app.go
davidterpay: otherwise invalid so imo its fine if we replace. its a bit of a footgun but i cant imagine why someone would do that (code would prevent this).

PR #626, opened 2024-07-25
feat: install script

technicallyty:
- adds an install script
- updates docs with curl command to execute remote script
- updates docs for the `From Source` option to checkout latest tag instead of using main


technicallyty:
Please help me test this!!!

run this:
```
curl -sSL https://raw.githubusercontent.com/skip-mev/slinky/tyler/blo-1525-curl-command-for-installing-connect/scripts/install.sh | sudo bash
```

aljo242:
works for me! 

do we want to call it `slinky` or `slinkyd`?

aljo242 (approved):

scripts/install.sh
```diff
@@ -0,0 +1,80 @@
...
+sudo mv "${SLINKY_BIN}" /usr/local/bin/
+
+# Make it executable
+sudo chmod +x /usr/local/bin/slinky
+
+# Clean up
+rm -rf "${TEMP_DIR}"
+rm "${FILE_NAME}"
+
+echo "Slinky ${VERSION} has been installed successfully!"
```
aljo242, line 80: could have another output that is just like `run slinky --help for more info` 

but not big deal really

codecov.yml
```diff
@@ -1,2 +1,3 @@
 ignore:
-  - ".github/**/*"
+  - "**/*.*"
+  - "!**/*.go"
```
Eric-Warehime: We should probably add some additional exclusions like `pulsar`, `pb`, etc.

PR #634, opened 2024-07-31
chore: Refactor utils into pure functions

Eric-Warehime:
Move these util functions out of the API handler so that they can be used as library functions in dydx.
Had to move the loggers into the args, and made it so they can be supplied as `nil` if you don't want to hook the zap logger up.


technicallyty (commented):
it seems like we only log on error messages, so alternatively you could just:

- return a more descriptive error (what the logger is spitting out currently)
- and change our internal callers of those functions to log the error

technicallyty:
oh nvm thats not true of all of them sadge.. maybe we could kill that log so the function sig is pretty 👉🏻👈🏻

aljo242 (approved):

technicallyty (approved):

PR #659, opened 2024-08-08
fix: error when oracle configuration has extra keys

wesl-ee:
Misconfiguration like this (missing the `api` key above `raydium_api`) will result in errors now when loading the overrides instead of quietly ignoring

```json
{
  "providers": {
    "raydium_api": {
      "endpoints": []
    }
  }
}
```


technicallyty (approved):
nice

PR #665, opened 2024-08-12
fix: clearer err msg when overriding an unknown provider type

wesl-ee:
Guards against typos like this in the oracle config:
```
{
  "providers": {
    "aydium_api": {
      "api": {
        "endpoints": [
            { ... }
        ]
      }
    }
  }
}
```

... errors now with `Error: failed to get oracle config: overridden key Providers[aydium_api].API.Timeout does not correspond to a known provider` instead of the vaguer `name cannot be empty`

### DIFF
--- a/cmd/slinky/config/config_test.go
+++ b/cmd/slinky/config/config_test.go
@@ -393,7 +393,7 @@ func TestReadOracleConfigWithOverrides(t *testing.T) {
 		tmpfile.Write([]byte(overrides))
 
 		_, err = cmdconfig.ReadOracleConfigWithOverrides(tmpfile.Name(), marketmap.Name)
-		require.Error(t, err)
+		require.ErrorContains(t, err, "overridden key")
 	})
 }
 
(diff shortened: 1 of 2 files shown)


aljo242:
@wesl-ee looks like some tests are failing

aljo242 (approved):

cmd/slinky/config/config_test.go
```diff
@@ -349,6 +349,52 @@ func TestReadOracleConfigWithOverrides(t *testing.T) {
...
+			prometheusServerOverride,
+			raydium.DefaultAPIConfig.Endpoints[0].URL,
+			endpointOverride.URL,
+			endpointOverride.Authentication.APIKey,
+			endpointOverride.Authentication.APIKeyHeader,
+		)
+		tmpfile.Write([]byte(overrides))
+
+		_, err = cmdconfig.ReadOracleConfigWithOverrides(tmpfile.Name(), marketmap.Name)
+		require.Error(t, err)
```
technicallyty: you could do something like `require.ErrorContains(t, err, "overridden key doesNotExist does not correspond to a known provider")` (or whatever unset[0] is)

just to make this test more confident that the error we expect is the one returned

technicallyty (approved):

wesl-ee:
This is only an error case for provider names we don't know about. If there was an error unmarshaling a provider config inside a provider value, I see a message like `Providers[raydium_api].API' has invalid keys: endpoint` which is helpful still

PR #677, opened 2024-08-16
feat: opt-out telemetry init

wesl-ee:
Pushes a mirror of the sidecar's prometheus metrics as opt-out telemetry to a remote statsd instance. Push address is 127.0.0.1:9125 currently, pending some infra to change this to a skip remote server. This is UDP so dropped packets mean nothing, the rules are made up and the points don't matter.


wesl-ee:
```
docker pull prom/statsd-exporter
docker run -d -p 9102:9102 -p 9125:9125 -p 9125:9125/udp                  prom/statsd-exporter
# ./build/slinky
```

localhost:9102 is then an endpoint that can be then scraped by prometheus

cmd/connect/config/config.go
```diff
@@ -22,6 +22,10 @@ const (
 	DefaultPrometheusServerAddress = "0.0.0.0:8002"
 	// DefaultMetricsEnabled is the default value for enabling prometheus metrics in slinky.
 	DefaultMetricsEnabled = true
+	// DefaultTelemetryDisabled is the default value for disabling telemetry.
+	DefaultTelemetryDisabled = false
+	// DefaultTelemetryDisabled is the default value for disabling telemetry.
```
Eric-Warehime:
```suggestion
	// DefaultTelemetryDisabled is the default value for the telemetry publication endpoint.
```

cmd/connect/main.go
```diff
@@ -192,6 +193,11 @@ func init() {
 		cmdconfig.DefaultMetricsEnabled,
 		"Enables the Oracle client metrics",
 	)
+	rootCmd.Flags().Bool(
+		flagTelemetryDisabled,
+		cmdconfig.DefaultTelemetryDisabled,
+		"Disabled the Oracle telemetry metrics",
```
Eric-Warehime:
```suggestion
		"Disables Oracle telemetry publication",
```

oracle/metrics/metrics.go
```diff
@@ -58,6 +67,11 @@ type Metrics interface {
 
 // OracleMetricsImpl is a Metrics implementation that does nothing.
 type OracleMetricsImpl struct {
+	promMetrics  *PromMetrics
+	statsdClient *statsd.Client
```
Eric-Warehime: Feels like this should be `statsd.ClientInterface` so that we can have a metrics impl that has nop statsd or has an actual client implementation.

Eric-Warehime: That would let us avoid all of the if branching around nil stasd clients in the method impls as well.

cmd/connect/config/config.go
aljo242: could we make this true?

oracle/metrics/metrics.go
```diff
@@ -71,113 +85,100 @@ type OracleMetricsImpl struct {
...
-			Name:      "health_check_market_providers",
+			Namespace: ProviderCountMetricName,
+			Name:      ProviderTickMetricName,
 			Help:      "Number of providers that were utilized to calculate the final price for a given market.",
 		}, []string{PairIDLabel}),
 		slinkyBuildInfo: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Namespace: OracleSubsystem,
-			Name:      "slinky_build_info",
+			Name:      SlinkyBuildInfoMetricName,
 			Help:      "Information about the slinky build",
```
aljo242: Can we make an issue to deprecate this metric and move to a `connect_build_info`

wesl-ee: done

oracle/metrics/metrics.go
wesl-ee: In my mind the nil check makes it clear this can be toggled on or off, but I see where this could be bad if we ever forget to check if it is nil and try to call that method

wesl-ee: So will change

aljo242:
@wesl-ee looks like lint failing and conflicts

wesl-ee:
jrRD7Uvn_400x400

aljo242 (approved):

Eric-Warehime (approved):

PR #696, opened 2024-08-23
refactor: remove unecessary nil checks in grpc handlers

technicallyty:


aljo242 (commented):
Callers could still call these message servers and pass `nil` requests.

There are cases where users could call the msg server during execution (therefore its not dispatched from the server natively).

I would rather keep these checks for safety.

PR #701, opened 2024-08-23
feat: namespace published metrics

wesl-ee:
Add the newly configured endpoint that receives statsd metrics and namespace metrics according to the moniker and network of the market map provider node. It does this over gRPC. In the case this information is not available (dial fails) data is simply not sent to the aggregator.


cmd/connect/config/config.go
```diff
@@ -220,3 +220,13 @@ func updateEndpointFromEnvironment(endpoint config.Endpoint, providerName string
 
 	return endpoint, endpointURL != nil || endpointAPIKey != nil || endpointAPIKeyHeader != nil
 }
+
+func GetMarketmapEndpointFromConfig(cfg config.OracleConfig) (config.Endpoint, error) {
+	for _, provider := range cfg.Providers {
+		if provider.Type == mmtypes.ConfigType {
```
Eric-Warehime, line 228: 🤔 Not sure this will work with [this](https://github.com/skip-mev/connect/blob/7bc5f1553e66a63125f4a356d56d57cfba98d202/cmd/constants/providers.go#L185) or [this](https://github.com/skip-mev/connect/blob/7bc5f1553e66a63125f4a356d56d57cfba98d202/cmd/constants/providers.go#L175) although the latter should go away soon.

oracle/metrics/node.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &NodeClient{
+		conn,
+	}, nil
+}
+
+func (nc *NodeClient) DeriveNodeIdentifier() (string, error) {
```
Eric-Warehime: Can probably write a simple unit test for this w/ a mock client.

wesl-ee:
rebased

wesl-ee:
rebase again

Eric-Warehime (approved):

PR #705, opened 2024-08-26
feat: block height staleness checker

aljo242:
Closes CON-1704

Add a `BlockAgeChecker` interface that defi providers can use when checking responses.


providers/apis/defi/types/block_age.go
```diff
@@ -0,0 +1,51 @@
...
+import "time"
+
+// BlockAgeChecker is a utility type to check if incoming block heights are validly updating.
+// If the block heights are not increasing and the time since the last update has exceeded
+// a configurable duration, this type will report that the updates are invalid.
+//
+//go:generate mockery --name BlockAgeChecker --output ./mocks/ --case underscore
+type BlockAgeChecker interface {
+	IsHeightValid(newHeight uint64) bool
+}
```
technicallyty: nit: why not just use the impl directly? object itself is simple enough, so an interface for it seems a bit unnecessary

aljo242: We can use a mock  version of it for testing

technicallyty: sure, but i dont see where this would ever be used. getting this to return true/false is easy enough as is. just set maxAge to 0 or something high for guaranteed success/fails

aljo242: @technicallyty removed interface

providers/apis/defi/ethmulticlient/multi_client.go
```diff
@@ -139,6 +149,7 @@ func (m *MultiRPCClient) BatchCallContext(ctx context.Context, batchElems []rpc.
 			r, ok := req[blockNumReqIndex].Result.(*string)
 			if !ok {
 				errs[i] = fmt.Errorf("result from eth_blockNumber was not a string")
+				results[i] = result{0, nil, errs[i]}
```
technicallyty: i think we can just remove the `errs` slice now, looks like its duplicate data as the error is getting put in `results` now

technicallyty (approved):
LGTM just some code that can be removed leftover

providers/apis/defi/types/mocks/block_age_checker.go
technicallyty, line 1: oh and this can be deleted

PR #712, opened 2024-08-28
feat: deprecate SlinkyBuildInfo

wesl-ee:
adds a ConnectBuildInfo metric and marks SlinkyBuildInfo as deprecated

### DIFF
--- a/oracle/metrics_test.go
+++ b/oracle/metrics_test.go
@@ -57,8 +57,8 @@ func (s *OracleTestSuite) TestMetrics() {
 		}
 	}()
 
-	metrics.On("SetSlinkyBuildInfo").Return()
-	metrics.On("AddTick").Return()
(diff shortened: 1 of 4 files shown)


aljo242 (approved):
LGTM

oracle/metrics_test.go
```diff
@@ -57,7 +57,7 @@ func (s *OracleTestSuite) TestMetrics() {
 		}
 	}()
 
-	metrics.On("SetSlinkyBuildInfo").Return()
+	metrics.On("SetConnectBuildInfo").Return()
```
technicallyty: if you want, the changes on main generates mocks with typesafe expects. so if you merge main and do

`make mocks`

you can change this line to:

```go
metrics.EXPECT().SetConnectBuildInfo().Return()
```

wesl-ee: excellent

PR #713, opened 2024-08-28
chore: typesafe mocks

technicallyty:
adds the expecter option to mockery. this allows you to use `EXPECT()` on mocks so you can call the method directly, instead of having to hardcode the method name. edited one test with an example.


aljo242 (approved):
LGTM  - just conflicts after we removed modules.

Can we make a task to move tests to use the typesafe versions?

technicallyty:
> LGTM - just conflicts after we removed modules.
> 
> Can we make a task to move tests to use the typesafe versions?

opened https://linear.app/skip/issue/CON-1718/update-mock-calls-to-use-expectmethodname-instead-of-onmethodname

PR #724, opened 2024-09-04
feat: validation service

aljo242:
Spins up and internal validation service that can introspect the live prices the sidecar is producing. 

If any of the prices is below our configured liveness threshold (99%, perhaps this should be stricter), the process will terminate itself. 

If all prices are valid after a configured amount of time, the process will gracefully tear down and end sucessfully.


pkg/http/address.go
```diff
@@ -0,0 +1,16 @@
+package http
+
+import "net"
+
+func IsValidAddress(address string) bool {
```
aljo242, line 5: moved to a common folder so it could be used by other code

contrib/compose/docker-compose-dev.yml
```diff
@@ -44,6 +44,7 @@ services:
       "--market-config-path", "/data/markets.json",
       "--pprof-port", "6060",
       "--run-pprof",
+      # "-m", "validate" uncomment to run in validation mode with default config
```
aljo242: how to run in validation mode

cmd/constants/marketmaps/markets.go
```diff
@@ -6354,67 +6259,6 @@ var (
 			  }
 			]
 		  },
-		  "RNDR/USD": {
```
aljo242, line 6357: removed the dead feeds.  Validation fails when they are here and passing when they are gone

cmd/connect/main.go
```diff
@@ -187,6 +200,20 @@ func init() {
...
+		"m",
+		string(modeExec),
+		"Select the mode to run the oracle in.  Default is \"exec\" which will fetch prices as configured.  \"validate\" mode will run the oracle for a set period of time to validate the configuration.",
+	)
+	rootCmd.Flags().IntVarP(
+		&validationPeriod,
+		flagValidationPeriod,
+		"",
+		int(validation.DefaultValidationPeriod),
+		"Duration to run in validation mode.  Note: this flag is only used if mode == \"validate\"",
```
Eric-Warehime: It's not clear to me what the units are for this int--hours minutes days etc.

contrib/compose/docker-compose-dev.yml
```diff
@@ -44,6 +44,7 @@ services:
       "--market-config-path", "/data/markets.json",
       "--pprof-port", "6060",
       "--run-pprof",
+      "-m", "validate" # uncomment to run in validation mode with default config
```
Eric-Warehime: Is this commented?

service/validation/validation.go
```diff
@@ -0,0 +1,106 @@
...
+			if len(missingPrices) > 0 {
+				v.logger.Warn("currently missing prices", zap.Any("prices", missingPrices))
+				for _, price := range missingPrices {
+					missingPricesMap[price]++
+				}
+			}
+		}
+	}()
+
+	time.Sleep(v.cfg.ValidationPeriod)
```
Eric-Warehime: Can you make this a select loop that checks the context periodically? I feel like graceful shutdown would be nice here especially if our validation period is long.

cmd/connect/main.go
technicallyty: there is `rootCmd.Flags().DurationVarP()`

Eric-Warehime (approved):

cmd/connect/main.go
```diff
@@ -75,12 +80,21 @@ var (
...
+	validationPeriod    time.Duration
 )
 
 const (
 	DefaultLegacyConfigPath = "./oracle.json"
 )
 
+type runMode string
+
+var (
```
technicallyty:
```suggestion
const (
```

cmd/connect/main.go
```diff
@@ -187,6 +201,19 @@ func init() {
...
+		flagMode,
+		"m",
+		string(modeExec),
+		"Select the mode to run the oracle in.  Default is \"exec\" which will fetch prices as configured.  \"validate\" mode will run the oracle for a set period of time to validate the configuration.",
+	)
+	rootCmd.Flags().DurationVar(
+		&validationPeriod,
+		flagValidationPeriod,
+		validation.DefaultValidationPeriod,
+		"Duration to run in validation mode in seconds.  Note: this flag is only used if mode == \"validate\"",
```
technicallyty: with the `DurationVar` change, 1h, 5m, etc is valid now :-)
```suggestion
		"Duration to run in validation mode.  Note: this flag is only used if mode == \"validate\"",
```

cmd/connect/main.go
```diff
@@ -395,6 +424,27 @@ func runOracle() error {
...
+
+		go func(c context.CancelFunc) {
+			defer c()
+
+			_, err := validatorService.Run(ctx)
+			if err != nil {
+				logger.Error("failed to validate metrics", zap.Error(err))
+
+				// kill the process
+				os.Exit(1)
```
technicallyty, line 441: careful with this if you still want the `defer c()` to run. deferred functions are not run when using os.Exit. not really sure if that changes any code paths or not

contrib/compose/docker-compose-dev.yml
technicallyty: should this be commented out? comment suggests so

service/validation/validation.go
```diff
@@ -0,0 +1,133 @@
...
+	"go.uber.org/zap"
+
+	oraclemetrics "github.com/skip-mev/connect/v2/oracle/metrics"
+)
+
+const (
+	DefaultNumChecks             = 1000
+	DefaultRequiredPriceLiveness = 99.0
+	DefaultValidationPeriod      = 10 * time.Minute
+	DefaultBurnInPeriod          = 60 * time.Second
```
technicallyty: the rest of these vars are pretty self explanatory, however  i think `BurnInPeriod` could use a comment

service/validation/validation.go
```diff
@@ -0,0 +1,133 @@
...
+	return Validator{
+		logger:  logger.Named("validation"),
+		metrics: metrics,
+		cfg:     cfg,
+	}
+}
+
+// Config includes information for configuring a validation service instance.
+type Config struct {
+	BurnInPeriod          time.Duration
```
technicallyty: or can comment here, either one works.

service/validation/validation.go
```diff
@@ -0,0 +1,133 @@
...
+// Run runs the validation service, checking for missing prices and accumulating liveness data.
+func (v *Validator) Run(ctx context.Context) (LivenessResults, error) {
+	v.logger.Info("running in validation mode", zap.Duration("validation period (s)",
+		v.cfg.ValidationPeriod))
+
+	v.logger.Info("waiting for burn in period to end", zap.Duration("period (s)", v.cfg.BurnInPeriod))
+	time.Sleep(v.cfg.BurnInPeriod)
+
+	missingPricesMap := make(map[string]int)
+	tickTime := v.cfg.ValidationPeriod / time.Duration(v.cfg.NumChecks)
```
technicallyty, line 92: after reading the config, i would have expected the ticker to tick on the duration i passed into `ValidationPeriod`. maybe im just not understanding the code here though.

technicallyty, line 92: should we validate `tickTime` being a good value? a `0` tickTime will panic

service/validation/validation.go
technicallyty, line 1: should we add a test for this logic?

cmd/connect/main.go
aljo242: This means the mm is invalid, so we just want to kill everything immediately.  Will never happen in prod running

PR #726, opened 2024-09-04
feat: Change GetPrice query to string

Eric-Warehime:
Change the GetPrice query to take a currency pair string instead of a CurrencyPair structure.

This will be a breaking change included in the connect v2 release.


proto/slinky/oracle/v1/query.proto
technicallyty, line 1: at what point do we create a v2 dir? seems like this would be it, but im not entirely sure of proto breaking process

aljo242:
@Eric-Warehime E2E tests failing because they used the older version of this query

Eric-Warehime:
Closing this in favor of #740

PR #736, opened 2024-09-10
feat: Reduce blast radius of errors in MarketMap

Eric-Warehime:
Adds a method to the marketmap type, `GetValidSubset`. This is used in the update lifecycle in order to be able to accept updates to the marketmap which do not fully pass ValidateBasic.

It will attempt to remove any `ProviderConfig` with an invalid `NormalizeByPair`, and then remove any `Market` which fails to pass `ValidateBasic`. 

If a market has some valid ProviderConfigs it can still run as long as it still satisfies the min provider count.


x/marketmap/types/market.go
```diff
@@ -40,6 +40,48 @@ func (mm *MarketMap) ValidateBasic() error {
 	return nil
 }
 
+// GetValidSubset outputs a MarketMap which contains the maximal valid subset of this MarketMap.
+//
+//	In particular, this will eliminate anything which would otherwise cause a failure in ValidateBasic.
+//	The resulting MarketMap should be able to pass ValidateBasic.
+func (mm *MarketMap) GetValidSubset() MarketMap {
```
aljo242: I feel like this function should return an `error` as well and call `Validate` to ensure that the returned map is valid

aljo242 (approved):
otherwise lgtm

x/marketmap/types/market.go
aljo242: There is a case where there is no valid subset and we should just return `nil, error`

Eric-Warehime: I think you're right--if the entire set of markets is invalid we should just keep everything the same. Returning an error will do that while I think returning an empty subset would just stop fetching any prices at all.

technicallyty (approved):
LGTM pending @aljo242's suggestion

PR #741, opened 2024-09-11
feat: Refactor partial failures

Eric-Warehime:
Adds additional logging around partial failures.
Also utilizes the `lastUpdated` field in the marketmap response in order to avoid having to do full marketmap comparisons. Reason motivating this change is both to short circuit the comparison logic and also that I've seen what looks like indeterminism in comparison when calculating valid subsets.


technicallyty (approved):

aljo242 (approved):
lgtm

PR #743, opened 2024-09-13
feat: add check forex markets to constants for checking

aljo242:


technicallyty (approved):
LGTM. do we want to be able to enable these with PR labels?

PR #754, opened 2024-09-17
feat: check if normalize markets are enabled in `marketmap`

aljo242:
Closes CON-1748


x/marketmap/keeper/keeper.go
```diff
@@ -196,13 +196,14 @@ func (k *Keeper) IsMarketValid(ctx sdk.Context, market types.Market) error {
 	// check that all markets already exist in the keeper store:
 	for _, providerConfig := range market.ProviderConfigs {
 		if providerConfig.NormalizeByPair != nil {
-			has, err := k.markets.Has(ctx, types.TickerString(providerConfig.NormalizeByPair.String()))
+			norm, err := k.markets.Get(ctx, types.TickerString(providerConfig.NormalizeByPair.String()))
 			if err != nil {
-				return err
+				return fmt.Errorf("unable to check if %s is normalized: %w", providerConfig.NormalizeByPair.String(), err)
```
technicallyty: nit: i think this would produce a confusing error

if we pass `FOO/USD` with NormalizeBy = `USDT/USD`, the error will say `unable to check if USDT/USD is normalized: <other_error>`

something like this might be more clear:

`"unable to get normalized market %s for %s", providerConfig.NormalizeByPair.String(), market.Ticker.String`

def should include the main market that caused this error since it's called in a loop on `ValidateState`, so we can pinpoint which market is problematic

aljo242: https://github.com/skip-mev/connect/pull/754/commits/57be32f517d334e41561903559b9f65f49d9b130

technicallyty (approved):

PR #767, opened 2024-09-20
fix: Fix Flaky Oracle Server Test

zrbecker:
This fixes the flaky oracle server test.

- Instead of creating a listener to get a random available port, I made it so we could just use that listener directly when creating the server.
- I added back the health check, the reason the heath check was causing issues before was that the listener was being closed as a defer call, so that make the port unavailable for the oracle server. However, using the listener directly solve this.
- I used /version instead of /prices for health check since /prices ends up calling the mock, which we have not set mock functions for.
- I changed the Timeout test to just check for an error instead of a specific error since the specific error it returns seems like an implementation detail of gRPC.

The test still seems a little flaky as I am getting this error after 20-30 runs

```
        server_test.go:215: 
                Error Trace:    /Users/zrbecker/code/skip/connect/service/servers/oracle/server_test.go:215
                Error:          Received unexpected error:
                                rpc error: code = DeadlineExceeded desc = context deadline exceeded
                Test:           TestServerTestSuite/TestOracleMarketMap
```

But it is certainly less flaky than it was.


aljo242:
@zrbecker looks like theres some conflicts

zrbecker:
> @zrbecker looks like theres some conflicts

should be resolved now

aljo242 (approved):

service/servers/oracle/server.go
```diff
@@ -80,13 +81,11 @@ func (os *OracleServer) routeRequest(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-// StartServer starts the oracle gRPC server on the given host and port. The server is killed on any errors from the listener, or if ctx is cancelled.
+// StartServer starts the oracle gRPC server with a given listener. The server is killed on any errors from the listener, or if ctx is cancelled.
```
technicallyty:
```suggestion
// StartServerWithListener starts the oracle gRPC server with a given listener. The server is killed on any errors from the listener, or if ctx is cancelled.
```

technicallyty (approved):

PR #770, opened 2024-09-23
docs: add quickstart guides for warden and initia

technicallyty:
- adds quickstart guides for warden and initia
- changes language around required version. we don't want people using versions above the specified anymore, since v2 has some breaking changes.


technicallyty:
@check-spelling-bot apply [updates](https://github.com/skip-mev/connect/actions/runs/11000602921/attempts/1).

docs/validators/quickstart.mdx
```diff
@@ -50,19 +50,29 @@ connect version
...
-    <Tab title="Stargaze Blockchain">
+    <Tab title="Stargaze">
         <QuickStartStargazeSnippet />
     </Tab>
+    <Tab title="Warden">
+        **The required Connect version for the Warden chain is `v1.0.10`.**
+        <QuickStartSnippet />
+    </Tab>
+    <Tab title="Initia">
+        **The required Connect version for the Initia chain is `v1.0.8`.**
```
aljo242: should we just bump these ?

technicallyty: like in their codebases? i was just going off https://github.com/initia-labs/initia/blob/main/go.mod#L49 and  https://github.com/warden-protocol/wardenprotocol/blob/main/go.mod#L58

aljo242: That's the protocol version, but the sidecar vals can run can be versioned separately right? 

Is there any reason not to have vals use the latest?

technicallyty: i don't think so, since connect v2 is proto breaking as we changed all the endpoints to be `connect/` instead of `slinky/` + all types/packages migrated

aljo242: But could we have them use `v1.0.12`?  This version includes the blast radius reduction included

technicallyty: probably, i'll run some tests and make sure it works on those chains

technicallyty: updated

aljo242 (approved):

PR #772, opened 2024-09-23
feat: removal message

aljo242:
Closes CON-1749

- adds notion of `ValidationHooks`
- adds `Options` configuration to keeper instantiation


x/marketmap/types/msg.go
```diff
@@ -124,3 +127,31 @@ func (m *MsgRemoveMarketAuthorities) ValidateBasic() error {
...
+	}
+
+	if len(m.Markets) == 0 {
+		return fmt.Errorf("markets to remove cannot be nil")
+	}
+
+	seenMarkets := make(map[string]struct{}, len(m.Markets))
+	for _, market := range m.Markets {
+		if _, seen := seenMarkets[market]; seen {
+			return fmt.Errorf("duplicate address %s found", market)
```
technicallyty:
```suggestion
			return fmt.Errorf("duplicate market %s found", market)
```

technicallyty (approved):

proto/connect/marketmap/v2/tx.proto
```diff
@@ -134,3 +139,20 @@ message MsgRemoveMarketAuthorities {
...
+  // Admin account.  This account is set in the module parameters.
+  string admin = 1 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+
+  // Markets is the list of markets to remove.
+  repeated string markets = 2;
+}
+
+// MsgRemoveMarketsResponse defines the
+// Msg/MsgRemoveMarketsResponse response type.
+message MsgRemoveMarketsResponse {}
```
zrbecker: I assume we are only succeeded if all markets to be removed exist, and are successfully removed which is why we don't need a response, maybe.

We could support making this message idempotent where we instead succeed if either the market does not exist, or was successfully removed. In that case it might make sense to return how many markets were actually removed.

x/marketmap/keeper/keeper.go
```diff
@@ -149,6 +148,19 @@ func (k *Keeper) UpdateMarket(ctx context.Context, market types.Market) error {
...
 }
 
+// DeleteDisabledMarket removes a Market if it is disabled.
+func (k *Keeper) DeleteDisabledMarket(ctx context.Context, tickerStr string) error {
+	// Check if Ticker exists
+	market, err := k.markets.Get(ctx, types.TickerString(tickerStr))
+	if err != nil {
+		return err
+	}
+	if market.Ticker.Enabled {
```
zrbecker: There is a bit of asymmetry here since we typically check for Enabled markets for updating in the ante handlers, but here are making it part of policy for all integrations.

Perhaps we should call this method "DeleteMarket", and write something like,

```
for _, handler := k.deleteMarketHandlers {
  if err := handler(ctx, market); err != nil {
    return err
  }
}
return k.markets.Remove(ctx, types.TickerString(tickerStr))
```

We can define a DefaultDeleteMarketHandlers() somewhere and use that to initialize the keeper.

Not sure if handler, verifier, validator is better.

This also ensures if we want to add more complex logic later for a specific integration, we can do so without changing connect.

x/marketmap/types/hooks.go
```diff
@@ -14,6 +14,9 @@ type MarketMapHooks interface {
 
 	// AfterMarketGenesis is called after x/marketmap init genesis.
 	AfterMarketGenesis(ctx sdk.Context, tickers map[string]Market) error
+
+	// AfterMarketRemoved is called after a market is removed.
+	AfterMarketRemoved(ctx sdk.Context, market Market) error
```
zrbecker, line 19: If this hook errors, does the message fail?

Could we check if the removed market was enabled here, and just fail?

zrbecker (approved):

proto/connect/marketmap/v2/tx.proto
aljo242: Updated to return `[]string` which is the array of markets that were successfully removed.

x/marketmap/keeper/keeper.go
aljo242: Added the notion of `ValidationHooks` which can be configured for the keeper.

Right now just passing them via options at instantiation which keeps the `NewKeeper` function signature the same

x/marketmap/types/hooks.go
aljo242: Yes if these hooks fail, the message would fail

zrbecker (approved):

PR #778, opened 2024-09-27
fix: clean up market map hooks 

aljo242:
Actually call the hook and clean it up to only use the ticker key as opposed to the whole market


technicallyty (approved):
is there a test we can add to make sure the hook is called as expected?

zrbecker (approved):

PR #785, opened 2024-10-02
feat: dynamic metrics

technicallyty:
adds a new metrics impl, `dynamic metrics` which is a thin wrapper for a Metrics impl. If the metrics impl is a  noop impl, and we have telemetry enabled, dynamic metrics will kick off a routine that tries to contact the node. if the node is contacted successfully, we attempt to switch our metrics impl to OracleMetricsImpl.


oracle/metrics/dynamic_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+		ticker := time.NewTicker(3 * time.Second)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-retryDuration.C:
+				return
+			case <-ticker.C:
+				// this is technically a race condition, but it doesn't really matter. if something is accessing the
```
zrbecker, line 72: What do you mean? Are you saying this is a race condition because the oracle would be using `d.impl` for reporting metric while we do this change here?

zrbecker (approved):

oracle/metrics/dynamic_metrics.go
technicallyty: correct

PR #797, opened 2024-10-16
fix: uniswap unable to handle more than 10 tickers

aljo242:
Closes CON-1849

Uniswap has been found to only support 10 tickers. Any ticker amount larger will cause the entire batch fetching to fail.

What I have found from investigating geth is that nodes are configured with a [batchItemLimit](https://github.com/search?q=repo%3Aethereum%2Fgo-ethereum%20batchItemLimit&type=code), but [default values](https://github.com/ethereum/go-ethereum/blob/15bf90ebc5eb8757b0a3e3fa3a5b6c8d3279c0be/node/defaults.go#L67) seem to be large enough for anything we realistically would do.

Regardless, from running the test I added and playing with the batchSize of our requests, all sizes > 10 will fail (any size LTE 10 works) for a variety of endpoints.
- batch of 1, passing (43 markets)
- batch of 5, passing (43 markets)
- batch of 10, passing (43 markets)
- batch of 11, failing (43 markets failing)
- batch of 43, failing (43 markets failing)

As it is written, the batch calls are serial, but we could parallelize this.  Until we are seeing realistic performance boundaries, I don't think we should prioritize this (and complicate the code).

Leaving the test and test file for demonstration.  Will remove the file and test once moving from draft.  The test file is generated for dydx main net, filtered for only uniswap markets


pkg/slices/slices.go
```diff
@@ -0,0 +1,20 @@
+package slices
+
+// Chunk chunks a slice into batches of chunkSize.
+// example: {1,2,3,4,5}, chunkSize = 2 -> {1,2}, {3,4}, {5}
+func Chunk[T any](input []T, chunkSize int) [][]T {
```
aljo242, line 5: taken from sauron. 

There is a [Chunk](https://pkg.go.dev/slices#Chunk) function in the `slices` pkg for go1.23, but this would require is to bump the go version in our go.mod, which would affect anyone using our code

providers/apis/defi/ethmulticlient/client.go
```diff
@@ -100,9 +100,9 @@ func (c *GoEthereumClientImpl) BatchCallContext(ctx context.Context, calls []rpc
 
 	if err = c.client.BatchCallContext(ctx, calls); err != nil {
 		c.apiMetrics.AddRPCStatusCode(c.api.Name, c.redactedURL, metrics.RPCCodeError)
-		return
+		return err
```
zrbecker, line 103: I assume this was done for clarity since it can be confusing when we have named return values. Consider removing the name entirely.

(i.e., instead of `(err error)` just let it be `error` for the return type.

providers/apis/defi/uniswapv3/fetcher.go
```diff
@@ -192,17 +193,23 @@ func (u *PriceFetcher) Fetch(
...
-		)
+	// process 10 tickers at a time
+	const batchSize = 10
+	batchChunks := slices.Chunk(batchElems, batchSize)
 
-		return types.NewPriceResponseWithErr(
-			tickers,
-			providertypes.NewErrorWithCode(err, providertypes.ErrorAPIGeneral),
-		)
+	for _, chunk := range batchChunks {
```
zrbecker, line 200: We should probably just parallelize this. It isn't too much more complicated, and this is one of the most common ways for latency to get out of control.

```go
	eg, egCtx := errgroup.WithContext(ctx)
	for _, chunk := range batchChunks {
		chunk := chunk
		eg.Go(func() error {
			// Batch call to the EVM.
			if err := u.client.BatchCallContext(egCtx, chunk); err != nil {
				u.logger.Debug(
					"failed to batch call to ethereum network for all tickers",
					zap.Error(err),
				)

				return err
			}

			return nil
		})
	}

	if err := eg.Wait(); err != nil {
		return types.NewPriceResponseWithErr(
			tickers,
			providertypes.NewErrorWithCode(err, providertypes.ErrorAPIGeneral),
		)
	}
```

providers/providertest/provider.go
```diff
@@ -111,7 +111,7 @@ func (c *Config) Validate() error {
 func DefaultProviderTestConfig() Config {
 	return Config{
 		TestDuration:   1 * time.Minute,
-		PollInterval:   5 * time.Second,
+		PollInterval:   20 * time.Second,
```
zrbecker: Any justification for changing the poll interval?

providers/providertest/provider.go
```diff
@@ -158,7 +158,7 @@ func (o *TestingOracle) RunMarketMap(ctx context.Context, mm mmtypes.MarketMap,
 			if len(prices) != expectedNumPrices {
 				return nil, fmt.Errorf("expected %d prices, got %d", expectedNumPrices, len(prices))
 			}
-			o.Logger.Info("provider prices", zap.Any("prices", prices))
```
zrbecker, line 166: Is this being removed because it was making the logs too messy?

providers/providertest/util.go
```diff
@@ -15,6 +15,9 @@ func FilterMarketMapToProviders(mm mmtypes.MarketMap) map[string]mmtypes.MarketM
 	for _, market := range mm.Markets {
 		// check each provider config
 		for _, pc := range market.ProviderConfigs {
+			// remove normalizations to isolate markets
+			pc.NormalizeByPair = nil
```
zrbecker, line 19: Is this just for testing, or why is it needed?

zrbecker (commented):
Left some comments without approval since this is still marked as draft. Feel free to request review whenever it's ready to go, or if you want additional feedback.

providers/providertest/util.go
aljo242: This is so that we can actually test the markets in isolation.

For example, if I want to isolate all of the Uniswap markets, they're all going to be Normalized by ETH/USD, which requires a bunch of other feeds to exist.

For this kind of testing, we only want to see "Does uniswap return a price for this expected pool?", so we just remove normalize pairs

providers/providertest/provider.go
aljo242: yep

providers/apis/defi/ethmulticlient/client.go
aljo242: https://github.com/skip-mev/connect/pull/797/commits/98c6a2b12b1805de36a769137fb6cc108acf488a

providers/providertest/provider.go
aljo242: reverted this, was just accidentally committed as I was experimenting

providers/apis/defi/uniswapv3/fetcher.go
zrbecker: If you do this, you should probably set a limit of something like 5 for the error group.

technicallyty (approved):
preapproving, pending @zrbecker suggestion of parallelizing the uniswap thing

zrbecker (approved):

PR #799, opened 2024-10-16
fix: last updated setting in `x/marketmap` keeper [DO NOT MERGE]

aljo242:
Closes CON-1839

- call `SetLastUpdated` for any `Set` or `Remove` calls in the keeper.  (previously these calls were only done during the MsgServer which meant that hooks, or any keeper calls that are not part of a MsgServer call were not updating LastUpdated)
- Add more testing around the `LastUpdated` field.  Use randomized blockheights instead of constants
- add checks to our E2E tests to ensure that the LastUpdated field is updated for all Tx we call.

Note, this fix is only for on-chain code, so a full upgrade would be needed.  This is also state-breaking, so we will need to make a new `v3.x` release as any `v1` or `v2` version of the code will result in disparate states (last updated will be updated differently).

I will make another PR that will only updates the sidecar, which can be rolled out more quickly.


tests/integration/connect_setup.go
```diff
@@ -332,16 +332,45 @@ func QueryMarket(chain *cosmos.CosmosChain, cp connecttypes.CurrencyPair) (mmtyp
...
+	cc, err := grpc.Dial(grpcAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, err
+	}
+	defer cc.Close()
+
+	// create the mm client
+	client := mmtypes.NewQueryClient(cc)
+
+	ctx := context.Background()
```
zrbecker: I know this is only used in tests, but can we just pass a ctx to this function and use it. Ideally a context is defined at the beginning of tests that we can use, and it looks like there is a context for us to use in this case, so we should avoid creating this context if possible.

tests/integration/connect_setup.go
```diff
@@ -413,16 +442,31 @@ func (s *ConnectIntegrationSuite) AddCurrencyPairs(chain *cosmos.CosmosChain, us
...
 	// broadcast the tx
-	resp, err := client.BroadcastTxCommit(context.Background(), tx)
+	txResp, err := client.BroadcastTxCommit(ctx, tx)
 	if err != nil {
 		return err
 	}
 
-	if resp.TxResult.Code != abcitypes.CodeTypeOK {
-		return fmt.Errorf(resp.TxResult.Log)
+	if txResp.TxResult.Code != abcitypes.CodeTypeOK {
```
zrbecker: Do proposals pass immediately with interchaintest?

x/marketmap/keeper/keeper.go
```diff
@@ -89,7 +89,18 @@ func (k *Keeper) GetMarket(ctx context.Context, tickerStr string) (types.Market,
 
 // setMarket sets a market.
 func (k *Keeper) setMarket(ctx context.Context, market types.Market) error {
-	return k.markets.Set(ctx, types.TickerString(market.Ticker.String()), market)
+	if err := k.markets.Set(ctx, types.TickerString(market.Ticker.String()), market); err != nil {
```
zrbecker, line 92: Just curious, what are the properties of this cosmos-sdk collections map? Does it handle stuff to ensure things are deterministic? Any idea of the specifics?

x/marketmap/keeper/msg_server.go
```diff
@@ -85,7 +85,7 @@ func (ms msgServer) UpsertMarkets(goCtx context.Context, msg *types.MsgUpsertMar
 		return nil, err
 	}
 
-	return &types.MsgUpsertMarketsResponse{}, ms.k.SetLastUpdated(ctx, uint64(ctx.BlockHeight())) //nolint:gosec
```
zrbecker, line 88: For these functions, SetLastUpdated will not get called many times per upsert potentially. Are there gas consequences of this? Is that a problem?

zrbecker (commented):
Added some comments. Let's chat about how this change and how it is integrated tomorrow.

x/marketmap/keeper/msg_server.go
aljo242: There are gas consequences because it is called many times yes.

This isn't a problem per-se as these transactions are only executed by gov / the mmu auth, but it is another reason why this PR is state breaking

tests/integration/connect_setup.go
aljo242: No, but the result of a proposal or submitted Tx will be TypeOK if it landed successfully

zrbecker (approved):

PR #800, opened 2024-10-16
fix: sidecar no longer checks `LastUpdated`

aljo242:
Part of CON-1839

This is a sidecar level fix that reverts the check of `LastUpdated` to checking equality of the incoming market map updates.

Factor out updating check to `IsMarketMapValidUpdated` which now has testing.

The market equality check can be reverted to the LastUpdated check once we are using on-chain logic with [this](https://github.com/skip-mev/connect/pull/799) PR.

### DIFF
--- a/oracle/market_mapper.go
+++ b/oracle/market_mapper.go
@@ -3,12 +3,61 @@ package oracle
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"strings"
 	"time"
 
 	"go.uber.org/zap"
+
+	mmtypes "github.com/skip-mev/connect/v2/x/marketmap/types"
 )
+
+// IsMarketMapValidUpdated checks if the given MarketMapResponse is an update to the existing MarketMap.
+// - returns an error if the market map is fully invalid or the response is invalid
+// - returns false if the market map is not updated
+// - returns true and the new market map if the new market map is updated and valid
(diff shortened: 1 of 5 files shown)


oracle/market_mapper.go
```diff
@@ -3,13 +3,61 @@ package oracle
...
+
+	mmtypes "github.com/skip-mev/connect/v2/x/marketmap/types"
 )
 
+// IsMarketMapValidUpdated checks if the given MarketMapResponse is an update to the existing MarketMap.
+// - returns an error if the market map is fully invalid or the response is invalid
+// - returns false if the market map is not updated
+// - returns true and the new market map if the new market map is updated and valid.
+func (o *OracleImpl) IsMarketMapValidUpdated(resp *mmtypes.MarketMapResponse) (mmtypes.MarketMap, bool, error) {
+	if resp == nil {
```
zrbecker, line 21: Looks like we didn't check for result.Value to be nil before, are we checking it now because we are not sure if it can be nil, or just because we are exposing this as a function?

zrbecker (approved):

oracle/market_mapper.go
aljo242: It should not be `nil` if it is returned from `x/marketmap`, but since the client is generic, its good to have nil checks in place since this is now a function

PR #807, opened 2024-10-24
chore: update codeowners to @skip-connect

technicallyty:
we've been bombarding wesley and eric with irrelevant github notifications lol

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,4 +4,4 @@
 
 # Primary repo maintainers
 
-*       @skip-connect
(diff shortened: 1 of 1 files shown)


zrbecker:
Would that @ actually work or do you need the teams tag?

zrbecker:
I think it needs to be this (check solve repo for reference)

* @skip-mev/skip-connect

technicallyty:
> I think it needs to be this (check solve repo for reference)
> 
> * @skip-mev/skip-connect

fixed 🤙🏻

zrbecker (approved):

PR #810, opened 2024-11-04
Prevent updates to decimals for enabled markets

chenyaoy:


zrbecker (commented):
We are going to reflect on this change a bit before accepting since in it's current form it would not be optional and if we needed to update decimals for any reason this prevents it completely. Perhaps it should be allow via governance update.

PR #834, opened 2024-11-21
chore: fix some comments

binchengqu:


aljo242:
we're winding down maintenance of Connect right now so we won't be merging this - apologies!

PR #836, opened 2024-11-21
docs: dydx v7

aljo242:
closes CON-1966

 wanted to keep the doc the same as people were running for dydx to keep the change that validators see as minimal as possible

### DIFF
--- a/docs/snippets/dydx-quickstart-guide.mdx
+++ b/docs/snippets/dydx-quickstart-guide.mdx
@@ -1,7 +1,155 @@
 <AccordionGroup>
+    <Accordion title="dYdX v7">
+
+        **The required version for Connect with dYdX v7 is `v1.0.13`.**
+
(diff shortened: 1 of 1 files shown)


docs/snippets/dydx-quickstart-guide.mdx
```diff
@@ -1,7 +1,155 @@
...
+            "endpoints": [
+        {
+            "url": "http://<YOUR NODE REST API ENDPOINT>"
+        },
+        {
+            "url": "<YOUR NODE GRPC ENDPOINT>:<YOUR NODE GRPC PORT>"
+        }
+            ]
+        }
+        },
```
technicallyty, line 28: given that v6 uses x/marketmap, i don't think users going from v6 -> v7 need to do this migration stuff anymore

aljo242: Yeah, I just wanted to keep the doc the same as people were running for dydx to keep the change that validators see as minimal as possible

technicallyty (approved):

PR #838, opened 2024-11-25
feat: Allow market authority to remove disabled markets

chenyaoy:
* Allow market authority to remove disabled markets
* Upgrade cosmossdk.io/math to fix vulnerability


aljo242:
ignoring but breaking as we are going to retract all previous versions

aljo242 (approved):

PR #865, opened 2024-12-11
chore: update reps and fix e2e

aljo242:
- update `mockery` to latest and generate mocks
- bump `interchaintest` and actually add the e2e tests (previously was removed)


zrbecker (approved):

tests/integration/slinky_integration_test.go
```diff
@@ -0,0 +1,130 @@
+package integration_test
```
aljo242, line 1: this file was not present meaning e2e was not actually  being run

PR #873, opened 2024-12-14
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


technicallyty:
We're winding down maintenance of Connect and are no longer accepting PRs. Thank you for your time.

PR #888, opened 2025-01-15
feat: add list query

aljo242:
- Add `Markets` query which returns a list from the state (rather than a map)
- Bump `cosmos-sdk` to fix vulnerability check


technicallyty (approved):
bigly

x/marketmap/keeper/keeper.go
```diff
@@ -135,6 +135,26 @@ func (k *Keeper) GetAllMarkets(ctx context.Context) (map[string]types.Market, er
 	return m, nil
 }
 
+// GetAllMarketsList returns the set of Market objects currently stored in state
+// as a list.
+func (k *Keeper) GetAllMarketsList(ctx context.Context) ([]types.Market, error) {
```
Eric-Warehime, line 140: Is this meant to be deterministically ordered? I'm not sure whether or not the iterator for collections will return the same order across different nodes. Also not sure whether it matters or not--seems like it might since the map that caused the issue had the same elements too right?

